Issue #1, opened 2020-05-21
null issues in config


moonmeister:
If I don't have a debug object in my config this throws `Cannot read property 'timeBuildSteps' of undefined`. Needs some optional chaining or config validation/defaults.

https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/06441416ee7917ac239eb338c21c69606454af13/src/utils/run-steps.js#L6

TylerBarnes:
Thanks @moonmeister! I think the default plugin options probably need to be deep merged since I think I'm probably just object merging. I'll get this fixed asap, thanks for the report!

TylerBarnes:
Just opened a PR to fix this, thanks again @moonmeister !

PR #3, opened 2020-05-22
fix: using all default plugin options breaks build

TylerBarnes:
fixes https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/issues/1

### DIFF
--- a/src/steps/process-and-validate-plugin-options.js
+++ b/src/steps/process-and-validate-plugin-options.js
@@ -29,9 +29,9 @@ const optionsProcessors = [
   {
     name: `queryDepth-is-not-a-positive-int`,
     test: ({ userPluginOptions }) =>
-      typeof userPluginOptions.schema.queryDepth !== `undefined` &&
-      (!isInteger(userPluginOptions.schema.queryDepth) ||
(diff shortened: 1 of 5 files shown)


TylerBarnes:
@moonmeister Turns out you were absolutely right about optional chaining here :) This step is before default plugin options are merged in as we're currently validating plugin options before merging them here. I fixed it and added an integration test to catch this in the future. Thanks again!

src/steps/process-and-validate-plugin-options.js
```diff
@@ -29,9 +29,9 @@ const optionsProcessors = [
   {
     name: `queryDepth-is-not-a-positive-int`,
     test: ({ userPluginOptions }) =>
-      typeof userPluginOptions.schema.queryDepth !== `undefined` &&
-      (!isInteger(userPluginOptions.schema.queryDepth) ||
-        userPluginOptions.schema.queryDepth <= 0),
+      typeof userPluginOptions?.schema?.queryDepth !== `undefined` &&
```
moonmeister, line 32: If you wanted to use optional chaining you can eliminate the type of undefined check i believe

PR #11, opened 2020-05-28
Fix: Use path prefix from helpers

trevorblades:
This branch updates the call to `fluid` by passing the `pathPrefix` option configured at the top level of a Gatsby configuration, rather than one from the plugin's own `pluginOptions`. This change means that if `gatsby build` is called without `--prefix-paths`, the `helpers.pathPrefix` will be an empty string, which is the correct behaviour.

### DIFF
--- a/src/steps/source-nodes/create-nodes/process-node.js
+++ b/src/steps/source-nodes/create-nodes/process-node.js
@@ -363,7 +363,7 @@ const replaceNodeHtmlImages = async ({
 
         const quality = pluginOptions?.html?.imageQuality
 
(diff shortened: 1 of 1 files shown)


TylerBarnes:
Hey

I just had a look and realized when I first added support for this I was looking at a regular gatbsy-image instead of an inline html gatsby-image 🤦 

This looks and works great

Thanks @trevorblades !

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@0.7.7`

PR #17, opened 2020-06-03
Feat: Docs!

TylerBarnes:
This PR adds decent docs which have been missing for a while. If you have time please read through and let me know what you think :)

Start at the main readme file https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/feat/docs/README.md


TylerBarnes:
@jasonbahl it'd be great if you could fact check this page https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/feat/docs/docs/problems-with-v3.md

docs/features/html-processing.md
```diff
@@ -0,0 +1,22 @@
...
+Media items in html are automatically sourced and image tags are swapped with `gatsby-image`'s
+
+Image tags in html that return 404's are logged to the console with a link to which post or page has the broken image. This allows you to easily discover and fix broken images that were deleted from the media library.
+
+This can be turned off with the `html.useGatsbyImage` boolean plugin option. See [plugin options](../plugin-options.md#html.usegatsbyimage-boolean) for more information.
+
+## Gatsby Link
+
+Anchor tag src's in html that are links to your WP instance are automatically rewritten to relative links (`https://yoursite.com/beautiful-page` becomes `/beautiful-page`). For this reason, it's recommended to use your WordPress page and post uri's to create your Gatsby page paths.
+
```
jacobarriola, line 16: @TylerBarnes perhaps we can provide an example (or point to one if it's already somewhere in these docs) of using your post/page uri when creating a page?

docs/problems-with-v3.md
```diff
@@ -0,0 +1,34 @@
...
+
+- The inflexibility and lack of an enforced schema for this API made proper node caching technically possible but functionally impossible.
+- The REST API is largely untyped, and REST does not enforce typing even if it was.
+  - This means Gatsby needed to use [inference](https://www.gatsbyjs.org/docs/glossary/#inference) to automatically build GraphQL types for the data that was returned from WP. The big issue with this is: if all data of a type is removed from WP, your Gatsby site's GraphQL queries will break because no data means Gatsby can no longer can infer the data structure. This led to a lot of odd and counterintuitive issues and a poor developer/content creator experience.
+  - If we were to manually write out types for the entire core WP REST API so that we no longer needed to use inference, any custom WP REST plugins would not work and would require a lot of additional tedious and error prone work to support these plugins changing their data structure over time.
```
jasonbahl, line 8: This also means that extensions, such as ACF to REST API can (and do) expose fields to WP REST endpoints that without being formally registered to the WP REST API Schema, meaning that the only way to know what's possible to be returned is to fetch every post and analyze what was returned. 

If the data coming from the WP REST API is different and unpredictable for each Post (or other resource), it makes it very difficult (impossible?) for decoupled applications to programmatically make use of the data being returned.

WPGraphQL resolves this by declaring what's possible to be returned as part of the Schema. Consumers, including Gatsby, can know what is possible via introspection of the GraphQL Schema without fetching a single Post.

moonmeister:
Looks really good! I can tell you spent a lot of time on writing and even little things like back links. :ok_hand: Nice Job. 

The only thing I have to add is related to downsides to `gatsby-source-graphql`. Because  it bypasses the Gatsby data layer you can't handle/modify node data with Gatsby APIs at build time. This means doing Image replacement, link handling, etc all has to be done client side. This adds bloat and pointless work in the browser.

0aveRyan:
@TylerBarnes this is awesome!

PR #43, opened 2020-07-09
Remove port from ngrok preview URL path

trevorblades:
This branch removes the `/8000` from the example ngrok preview URLs.


TylerBarnes:
Thanks Trevor!

Issue #55, opened 2020-07-14
Feat: Add error message about pluginOptions.schema.perPage for timeouts


TylerBarnes:
Right now when a request times out the error message mentions increasing the timeout, but another possibility is decreasing the perPage setting to lower the amount of work the remote server needs to do for each request. The error message in `src/utils/fetch-graphql.js` on line 207 should mention this.

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@1.0.14` Thanks again @jacobarriola !

PR #86, opened 2020-07-23
Include message about schema.perPage on timeout

jacobarriola:
Related to #55

### DIFF
--- a/src/utils/fetch-graphql.js
+++ b/src/utils/fetch-graphql.js
@@ -206,7 +206,7 @@ const handleFetchErrors = async ({
       formatLogMessage(
         `It took too long for ${url} to respond (longer than ${
           timeout / 1000
-        } seconds). Either your URL is wrong, you need to increase server resources, you need to decrease your options.schema.perPage value from the default 100, or you need to increase your timeout in the gatsby-source-wordpress options. \n${genericError(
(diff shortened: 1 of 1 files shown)


src/utils/fetch-graphql.js
```diff
@@ -206,7 +206,7 @@ const handleFetchErrors = async ({
       formatLogMessage(
         `It took too long for ${url} to respond (longer than ${
           timeout / 1000
-        } seconds). Either your URL is wrong, you need to increase server resources, or you need to increase your timeout in the gatsby-source-wordpress options. \n${genericError(
+        } seconds). Either your URL is wrong, you need to increase server resources, you need to decrease your options.schema.perPage value from the default 100, or you need to increase your timeout in the gatsby-source-wordpress options. \n${genericError(
```
TylerBarnes, line 209: Hey Jacob, thanks for this! I have a feeling it gets hard to read when the list of things grows beyond 3. What do you think about something like this?
```suggestion
        } seconds). Either your URL is wrong, you need to increase server resources, or you need to decrease the amount of resources each request takes. You can configure how much resources each request takes by lowering your `options.schema.perPage` value from the default of 100 nodes per request. Alternatively you can increase the request timeout by setting a value in milliseconds to `options.schema.timeout`, the current setting is ${timeout}. \n${genericError(
```

jacobarriola: Oh yeah, that works!

TylerBarnes:
Sweet! Thanks Jacob! 😄

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@1.0.14` 🚀

PR #90, opened 2020-07-24
mention which HTTP method is used for Webhook requests

### LINKED ISSUE #80: Docs: "Configuring WPGatsby - Setting Up Builds" doesn't mention which http method is used
It would be really helpful if the docs explicitly mentioned that the webhook makes a **POST** request to the given url.

vscncls:
closes #80


TylerBarnes:
Awesome :D thanks so much @vscncls !!

Issue #95, opened 2020-07-27
Search and replace for links


philmuze:
Hi guys, really great work with this new version of the plugin. This pairing with the wpgraphl-plugin is awesome. I'm currently working on a migration to v4 for my project. One thing I used there a lot was the search and replace functionality for links - see [here](https://github.com/gatsbyjs/gatsby/pull/3498/files/cfe0d9d48fc5f29053e91b80c1e20e6c2f62924f). Do you have any plans to integrate this also to the version v4?

TylerBarnes:
Hi @philmuze , thanks for opening this issue. The plugin currently auto-replaces any links to the frontend of the WP site. `https://your-wp-site.com/sample-page` is automatically rewritten to `/sample-page`. These links are then automatically caught browser-side and cause a client navigation when clicked instead of a full reload. See this docs page for more info https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/docs/features/html-processing.md#gatsby-link

Is there another use-case that this doesn't cover? I haven't added a search/replace for this reason but I might have missed something

philmuze:
We used this to serve our images from the CDN. Our backend is hosted with Pagely and they have this nice CDN feature for media files, which will be automatically uploaded to the CDN. The search and replace option was pretty straight forward to serve the images from the CDN. We skipped the gatsby image integration in this case.

TylerBarnes:
Interesting, I can see this being valuable for sure. Thanks for elaborating!

TylerBarnes:
I've added this to my todo list but if there's anyone that'd like to help out check this line https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/src/steps/source-nodes/create-nodes/process-node.js#L613

Here we have a list of "nodeStringFilters". We need to add a third one "findAndReplace". Within this new filter we should check plugin options for `options.findAndReplace`. This should be an array of objects something like:
```json
{
  options: {
    findAndReplace: [
      {
        find: ``, // regexp
        replace: `` // string
      }
    ]
  }
}
```

The reason this should be an array is there may be multiple things that need to be replaced (maybe). Since it's not much more effort or maintenance burden to allow multiple replaces we should do that.

philmuze:
Hi @TylerBarnes could you have a look for my PR in this topic #129.

TylerBarnes:
[...]

PR #129, opened 2020-08-17
Add plugin option search and replace

philmuze:
Hi 👋 ,

during my upgrade gatsby-source-wordpress-experimental I missed the search and replace option from the previous version. We use this for example to replace our basic Wordpress URL with a CDN URL to serve the images from there. So with this PR I added the option to the new plugin version as well.

This solves #95.


acao:
so you're referring to this functionality in the old source plugin, correct?

```js
// Search and Replace Urls across WordPress content.
searchAndReplaceContentUrls: {
  sourceUrl: "https://source-url.com",
  replacementUrl: "https://replacement-url.com",
},
```

I like how your implementation expands on this

src/steps/source-nodes/create-nodes/process-node.js
```diff
@@ -799,6 +799,35 @@ const replaceNodeHtmlLinks = ({ wpUrl, nodeString, node }) => {
   return nodeString
 }
 
+// replaces specific string with a given string from the plugin options config
+const findAndReplaceNodeHtmlString = ({ nodeString, node, pluginOptions }) => {
+  if (pluginOptions?.findAndReplace?.length) {
```
acao, line 804: I think we should be looking for `pluginOptions?.searchAndReplace` according to the docs.

also, this would be a great place to check `Array.isArray(pluginOptions?.searchAndReplace)`. However, now that we have plugin options validation, duck typing checks like these won't be as important

acao (changes requested):
Looking awesome all around. Do you mind if we finish it up for you?

Let's just make sure the plugin option names and doc names match up, and what's more, we just added plugin option validation, so when you pull in the upstream you can add joi validation for this option as well!

philmuze:
@acao thanks for the review. Add the suggested change. But to be honest I never worked with this joi validation. Do you have any tips here?

munyah:
erm, sorry, excuse the interruption, what happened to this feature @philmuze?

TylerBarnes:
Sorry @philmuze and @munyah this PR slipped between the cracks during some internal changes at Gatsby. If you're able to open this PR on the Gatsby repo for `gatsby-source-wordpress` I'm happy to review/merge it. Thanks everyone!

PR #158, opened 2020-09-17
fix nodeMetaByUrl caching issue

### LINKED ISSUE #153: LocalFile cache confusion
Unfortunately I cannot share my wordpress instance, however let me explain my findings.

I have multiple pages that reference the same image in different sizes.

in `fetchNodeHtmlImageMediaItemNodes` we can see that the `imageNode.id` is saved to the cache

```
    if (imageNode) {
      // save any fetched media items in our global media item cache
      _store.default.dispatch.imageNodes.pushNodeMeta({
        sourceUrl: htmlImgSrc,
        id: imageNode.id,
        modifiedGmt: imageNode.modifiedGmt
      }); // match is the html string of the img tag
```

In my case the `imageNode` is a `MediaItem`
```
  ...
  "remoteFile": {
    "id": "f61fe9db-337d-5f25-8ac0-35a9a8a9f501"
  },
  "localFile": {
    "id": "f61fe9db-337d-5f25-8ac0-35a9a8a9f501"
  },
  "parent": null,
  "internal": {
    "contentDigest": "2001149f43ce77ed8758db8595e7e8fd",
    "type": "WpMediaItem",
    "counter": 77,
    "owner": "gatsby-source-wordpress-experimental"
  },
  "type": "MediaItem",
  "nodeType": "MediaItem",
  "children": []
```

However the `pushNodeMeta` call also causes the `nodeMetaByUrl` lookup table to be set / overwritten.

```
    pushNodeMeta(state, {
      id,
      sourceUrl,
      modifiedGmt
    }) {
      state.nodeIds.push(id);
      state.nodeMetaByUrl[(0, _fetchReferencedMediaItems.stripImageSizesFromUrl)(sourceUrl)] = {
        id,
        modifiedGmt
      };
      return state;
    }
```

However in `getFileNodeByMediaItemNode` we pull the node meta out and treat it as if it was a localFile

```
const getFileNodeByMediaItemNode = async ({
  mediaItemNode,
  helpers
}) => {
  const {
    sourceUrl,
    modifiedGmt,
    mediaItemUrl,
    databaseId
  } = mediaItemNode;
  const fileUrl = sourceUrl || mediaItemUrl;

  if (!fileUrl) {
    helpers.reporter.warn((0, _formatLogMessage.formatLogMessage)(`Couldn't find source url for media item #${databaseId}`));
    return null;
  }

===>   const existingNodeMeta = getFileNodeMetaBySourceUrl(fileUrl);

  if ( // if we already have this image
  existingNodeMeta && existingNodeMeta.id && // and it hasn't been modified
  existingNodeMeta.modifiedGmt === modifiedGmt) {
    const node = await helpers.getNode(existingNodeMeta.id);
    return node;
  }

  return null;
};
```

So in my case the `pushNodeMeta` gets called multiple times, the first few instances with the correct remoteFileNode `id` and `url` coming from `createRemoteMediaItemNode`
```
 // so we can touch this node next time
  // and so we can easily access the id by source url later

  _store.default.dispatch.imageNodes.pushNodeMeta({
    id: remoteFileNode.id,
    sourceUrl: mediaItemUrl,
    modifiedGmt
  });
```

However, a bit later we get the call with a resized version of the url (coming from `fetchNodeHtmlImageMediaItemNodes`). Now `pushNodeMeta` gets called with the id of the imageNode.  Since both urls are normalized (`stripImageSizesFromUrl`) the cache in `nodeMetaByUrl` gets overwritten and it now points to an `ImageNode` rather than a `localFile`.

Now every call to `getFileNodeByMediaItemNode` will result in the ImageNode being returned rather than the FileNode.

In the end the build fails with 

```
The "path" argument must be one of type string, Buffer, or URL. Received type undefined



  Error: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be one of ty  pe string, Buffer, or URL. Received type undefined
  ```

called from `replaceNodeHtmlImages` when going through `htmlMatchesWithImageResizes` because

```
      if (imageResize) {
        imgOptions.fluid = imageResize
        ReactGatsbyImage = React.createElement(Img, imgOptions, null)
      } else {
        const { fileNode } = matchResize

        const fileName = `${fileNode.internal.contentDigest}/${fileNode.base}`
```

the `fileNode` here is no longer a real `fileNode` but the `imageNode`.

rburgst:
- in the case where the same image gets referenced both in an ACF (without size spec) and within HTML
  then getFileNodeByMediaItemNode
  will return a medianode rather than
  the expected file node
- this patch avoids overwriting the
  `nodeMetaByUrl` cache
  as well as ensuring that
  `getFileNodeByMediaItemNode` will
  always return a fileNode

- note that I am not certain that both changes are strictly needed 

- fixes #153

### DIFF
--- a/src/models/image-nodes.js
+++ b/src/models/image-nodes.js
@@ -24,9 +24,14 @@ const imageNodes = {
 
     pushNodeMeta(state, { id, sourceUrl, modifiedGmt }) {
       state.nodeIds.push(id)
-      state.nodeMetaByUrl[stripImageSizesFromUrl(sourceUrl)] = {
-        id,
-        modifiedGmt,
+      const nodeUrl = stripImageSizesFromUrl(sourceUrl);
+      // dont overwrite the lookup table in case we have multiple
+      // sized urls for the same image
+      if (!state.nodeMetaByUrl[nodeUrl]) {
+        state.nodeMetaByUrl[nodeUrl] = {
+          id,
+          modifiedGmt,
+        };
       }
 
       return state
--- a/src/steps/source-nodes/create-nodes/create-remote-media-item-node.js
+++ b/src/steps/source-nodes/create-nodes/create-remote-media-item-node.js
@@ -72,9 +72,20 @@ export const getFileNodeByMediaItemNode = async ({
     // and it hasn't been modified
     existingNodeMeta.modifiedGmt === modifiedGmt
   ) {
-    const node = await helpers.getNode(existingNodeMeta.id)
+    let node = await helpers.getNode(existingNodeMeta.id)
 
-    return node
+    // some of the cached node metas dont necessarily need to be a File
+    // so check that we really read a file node
+    if (node.internal.type !== 'File') {
+      if (node.localFile && node.localFile.id) {
+        // look up the corresponding file node
+        node = await helpers.getNode(node.localFile.id)
+      } else {
+        return null;
+      }
+    }
+
+    return node;
   }
 
   return null


TylerBarnes:
Published in `gatsby-source-wordpress-experimental@1.4.2`

PR #159, opened 2020-09-17
allow alternative urls for images

rburgst:
- sometimes you clone your site (e.g. staging env) and it uses a different root url
- if you simply copy the db content then image urls are not rewritten to the new domain by wordpress
- one way to solve this would be to simply use a wordpress plugin to rewrite the urls
- another way is to rewrite the urls within this plugin


TylerBarnes:
Hey, thanks for this, it looks really useful! I want to spend some more time on this but need to ship the stable release of this plugin first. Looking forward to checking it out, thanks again!

PR #162, opened 2020-09-17
Feat/int test gutenberg

### LINKED ISSUE #149: blocksJSON are messing with useGatsbyImage
Hi all,

I am using Gutenberg as the main editor and I have lots of different innerBlocks with different attributes and it is getting too overwhelming with direct querying or fragments. I found a WP plugin [WP GraphQL Gutenberg](https://github.com/pristas-peter/wp-graphql-gutenberg) that enables me to expose Gutenberg blocks and also expose them as a stringified JSON, which is super useful for me.

However, when I am sourcing data through your plugin and have the useGatsbyImage API set to true, it returns me an error, because it also goes through the url parser (error right below).

```
url passed to create-remote-file-node is either missing or not a proper web uri: %5C%22https:%5C/%5C/...
```

I tried adding .replace with regex for backslash and double-quotes, which works, but breaks the stringified JSON response (no escaping of double quotes now).
Falsing useGatsbyImage lets me build the site, but then I cannot query the media within the blocks.
Excluding blocksJSON field name lets me build the site, but I would have to write all the fragments.

### How to reproduce
Use WP site with Gutenberg, use WP GraphQL Gutenberg plugin together with other necessary WP GraphQL plugins, use gatsby-source-wordpress-experimental in your Gatsby project with enabled useGatsbyImage, you should get an error.

### What is expected
There can be more options:
1. This plugin would skip URLs it doesn't understand (e.g. these from blocksJSON with backslashes) not return an error.
2. This plugin would clean these URLs first and would return the gatsby-image block, but would not break the stringified JSON if that's possible
3. This plugin would skip the URLs from  element and would look for blocks attributes.url or something similar

I am using latest versions of all plugins and Gatsby.

TylerBarnes:
This PR adds resolver integrity integration tests by making similar GraphQL requests to Gatsby & WPGraphQL and comparing the results. wp-graphql-gutenberg is used as a testing schema here to help cover more types of fields and also ensure wp-graphql-gutenberg works properly with this plugin.

Along the way I found and solved 2 bugs:

- Fields that are encoded JSON were being parsed for Gatsby images which shouldn't be happening (so this PR fixes #149)
- The WPGQL `Node` type is an interface type of all possible WPGQL node types. Since not all of those are Gatsby node types, fields of this type were not being registered properly as Gatsby node connections. This PR adds a check to the default resolver to see if a connected node exists as well as adding the `__typename` field to all interface fields during query generation and node sourcing.


TylerBarnes:
Published in `gatsby-source-wordpress-experimental@1.3.9`

PR #182, opened 2020-09-28
fix undefined dereference in case the cached node no longer exists

rburgst:
- in the case where the cached node is no longer in the node tree
  (possibly due to a refresh)
  then we were trying to dereference `node.internal` when
  `node` was `undefined`


TylerBarnes (approved):
Thanks!

Issue #197, opened 2020-10-02
Bug: Source and transform nodes step never finishes after 1.4.1


gustavpursche:
After version `1.4.1` I run into the following problem:

```
kolle-on-xps :: Development/coaltransitions » yarn develop
yarn run v1.22.4
$ gatsby develop
success open and validate gatsby-configs - 0.066s
success load plugins - 0.701s
success onPreInit - 0.032s
info One or more of your plugins have changed since the last time you ran Gatsby. As
a precaution, we're deleting your site's cache to ensure there's no stale data.
success initialize cache - 0.074s
success copy gatsby files - 0.017s
success onPreBootstrap - 0.015s
success  gatsby-source-wordpress  ingest WPGraphQL schema - 1.276s
success createSchemaCustomization - 2.189s
success  gatsby-source-wordpress  fetch root fields - 1.591s
success  gatsby-source-wordpress  Category - 1.682s - fetched 1
success  gatsby-source-wordpress  Comment - 1.796s - fetched 0
success  gatsby-source-wordpress  Menu - 1.787s - fetched 2
success  gatsby-source-wordpress  Finding - 2.669s - fetched 8
success  gatsby-source-wordpress  NewsEntry - 2.942s - fetched 3
success  gatsby-source-wordpress  Post - 3.261s - fetched 1
success  gatsby-source-wordpress  MenuItem - 3.484s - fetched 8
success  gatsby-source-wordpress  PostFormat - 3.777s - fetched 0
success  gatsby-source-wordpress  ContentType - 3.892s - fetched 8
success  gatsby-source-wordpress  UserRole - 4.695s - fetched 0
success  gatsby-source-wordpress  Taxonomy - 5.137s - fetched 3
success  gatsby-source-wordpress  ResearchProject - 5.527s - fetched 17
success  gatsby-source-wordpress  User - 6.000s - fetched 4
success  gatsby-source-wordpress  Page - 6.208s - fetched 13
success  gatsby-source-wordpress  Researcher - 8.610s - fetched 62
success  gatsby-source-wordpress  Tag - 11.001s - fetched 157
success  gatsby-source-wordpress  Publication - 11.071s - fetched 110
success Downloading remote files - 2.055s - 0/2 0.97/s
info pushing https://admin.coaltransitions.org/wp-content/uploads/2019/10/diw-weekly-report-41-2.png to the end of the request queue.
info Previously retried 2 times already.
[...]

Issue #53, opened 2020-07-14
Fetching images at build time causes an error


shosasaki0:
Hi

I'm about to try out a great plugin, but when I push the first image, it says An error has occurred and I am unable to proceed.

The error is this.

```
already re-queued /wp-content/uploads/2020/06/test.jpg 5 times :( sorry.
Try lowering process.env.GATSBY_CONCURRENT_DOWNLOAD.
It's currently set to undefined
```
It's not just this image that is showing an error.

The image exists and it can be seen when I hit the URL.
The images are all over 1MB in size, is there a reason for this?
Changing GATSBY_CONCURRENT_DOWNLOAD did not solve the problem It was.

It probably stops here.
```
source and transform nodes
gatsby-source-wordpress  fetching nodes — 258 total
gatsby-source-wordpress  creating nodes
gatsby-source-wordpress  MediaItem — fetched 200
```

Thank you for all the work on this plugin!

Regards,
sasaki

TylerBarnes:
Hi @shosasaki0 , thanks for opening an issue!
It's hard to tell what it could be from your description. Are you able to share a reproduction repo and live WPGraphQL API endpoint that I can use to debug?

Thanks!

lupoair:
@TylerBarnes We are having the same issue with out of the box gatsy + gatsby-source-wordpress-experimental setup up to our live endpoint. Can send endpoint for testing as PM.

TylerBarnes:
Hi @lupoair , it's really hard to keep track of which GraphQL endpoints map to which GH issues. Are you able to send your API again?

BTW I think this issue might be resolved now. It looks like the url it's trying to fetch is a relative url `/wp-content/uploads/2020/06/test.jpg`. The plugin more recently detects this and adds the WP url to the request.

kawamataryo:
Hi. I have the same problem when `gatsby develop` commands.
Is there a solution?

## Environment to reproduce
Repo: https://github.com/kawamataryo/practice-gatsby-source-wordpress-experimental

Wordpress: v5.51
WP Gatsby: 0.4.16
WP GraphQL: 0.12.3

### log

```
~/g/g/k/practice-gatsby-source-wordpress-experimental ❯❯❯ gatsby develop
success open and validate gatsby-configs - 0.037s
success load plugins - 1.697s
success onPreInit - 0.034s
success initialize cache - 0.012s
success copy gatsby files - 0.044s
success onPreBootstrap - 0.019s
[...]

PR #198, opened 2020-10-04
Fix/failed fetching images to build

shosasaki0:
#53 #197
In version 1.5.3, an error occurred at the part where the first image was fetched and I could not proceed.
This seems to be limited to sites that use basic authentication, and according to the document, the auth option is deprecated for fetching by got.
https://github.com/sindresorhus/got#options
```
Note: Legacy URL support is disabled. options.path is supported only for backwards compatibility. Use options.pathname and options.searchParams instead. options.auth has been replaced with options.username &  #options.password.
```
Changed to accommodate it :)

### DIFF
--- a/src/steps/source-nodes/create-nodes/create-remote-file-node/index.js
+++ b/src/steps/source-nodes/create-nodes/create-remote-file-node/index.js
@@ -249,7 +249,8 @@ async function processRemoteNode({
   // extensible. We should define a proper API that we validate.
   const httpOpts = {}
   if (auth && (auth.htaccess_pass || auth.htaccess_user)) {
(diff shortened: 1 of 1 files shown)


TylerBarnes:
Woo! I was able to reproduce on node 14.13.1 and your PR fixes it! Your PR also still works on node 10.19.1 👍 

Thanks for this!!

jdselig:
Woo, thanks again guys! 🙌

TylerBarnes:
Hmm, looks like I'm running into some backwards compatibility issues actually. Working on it now though 👍

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@1.6.2`! Thanks for checking this out everyone!

Issue #131, opened 2020-08-18
Are PDF's or MP4's retrieved from wordpress media items? 


ssoney200:
Hi guys, 

All my jpegs and pngs etc are downloaded using Gatsby source wordpress experimental but I also want to pull mp4 videos and PDF files from the media in Wordpress.

I'm sure this was working when I used the Rest API version, but I can't locate them anywhere using this version.

Am I missing something? 

Thanks,
Jamie

philip-akerfeldt:
I also have this problem. 
Images are fetched fine from WPGraphql using the `WpMediaItem` type. PDFs however, are missing some important fields. For me its the `sourceUrl` that is the biggest issue right now.

moonmeister:
Do either of you have some code I can see or more specific logs? 

Some possibilities: 

1. non image items in WordPress don't have `sourceUrl` the use `mediaItemUrl` if memory serves (images also have `meduaItemUrl` so it's probably better to use that).  
2. I believe images are only downloaded into Gatsby if they are referenced by something in WP. meaning they're in a post/page, linked in a menu, etc. If it's just free floating in the media library it's not accessible for now from Gatsby.

philip-akerfeldt:
Sure! Here are som additional information about my case.
I am creating pages for my PDF files so my initial query that is used in gatsby-node is the following:
```    
{
        allWpMediaItem(filter: {mimeType: {regex: "/(pdf)/"}}) {
            nodes {
                id
                slug
                mimeType
                type: nodeType
                status
            }
        }
    }
```

On each specific PDF page I am using the following fragment query:
```
fragment singleWpPDFFragment on WpMediaItem {
    id
    slug
    sourceUrl
    status
    localFile {
      prettySize
      name
      ext
      publicURL
    }
  }
```

The error I get right now does not seem to be connected to the mediaItem type. Ill get back to you with more information when I get the time to debug and test. Should be in a couple of hours.

Im gonna try your first suggestion @moonmeister! 
This is a huge issue for my project that is highly reliant on having PDFs as their own pages 😬 
I will post an update shortly 😄

ssoney200:
> Sure! Here are som additional information about my case.
> I am creating pages for my PDF files so my initial query that is used in gatsby-node is the following:
> 
> ```
> {
>         allWpMediaItem(filter: {mimeType: {regex: "/(pdf)/"}}) {
>             nodes {
>                 id
>                 slug
>                 mimeType
[...]

PR #202, opened 2020-10-07
Feat: process non-image files in html

TylerBarnes:
### New Features

- A new plugin option `options.html.createStaticFiles` was added. When this is `true`, any url's which are wrapped in "", '', or () and which contain `/wp-content/uploads` will be transformed into static files and the url's will be rewritten. This adds support for <audio>, <video>, and <a> tags which point at WP media item uploads as well as inline-html css like background-image: url(). It will also transform any plain text like "https://yoursite.com/wp-content/uploads/image.png" as long as it's wrapped in "", '', or ().

### Changes

- Connected media item id's are analyzed in non media item nodes to determine which nodes to fetch. That was previously being done by finding connections that had an `id` and a `sourceUrl` field on the fetched node data. This was switched to use `id` and `__typename` to save bytes across the wire, slightly reduce the number of db lookups that need to be done on the WPGQL side during node sourcing, and prevent the `createStaticFiles` option from detecting and transforming `sourceUrl` fields.


TylerBarnes:
Related to #131

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@1.6.0`

Issue #70, opened 2020-07-15
Explain in error messages why "404'ing Media causes build to fail"


ashhitch:
Previously I would just get a warning of 404s for media but now when the first post with inserted media which 404s I get an error and the build fails.

```
Received a 404 when trying to fetch

https://api.thesite.co.uk/wp-content/uploads/2013/11/lounge-diner-300x300.jpeg
from Post #4111 "Post with Image 404"

Most likely this image was uploaded to this Post and then deleted from the media library.
You'll need to fix this and re-save this Post to remove this warning at
https://api.thesite.co.uk/wp-admin/post.php?post=4111&action=edit.

 ERROR

Cannot read property 'id' of null

  TypeError: Cannot read property 'id' of null

  - process-node.js:196 fetchNodeHtmlImageMediaItemNodes
    [gatsby-theme-starter]/[gatsby-source-wordpress-experimental]/steps/source-nodes/create-nodes/process-node.js:196:23

  - next_tick.js:68 process._tickCallback
    internal/process/next_tick.js:68:7
```

The build completes if I set:

```
 html: {
   useGatsbyImage: false,
},
```

Plugin version `1.0.7`
WP Gatsby `0.4.9`
WpGraphQL `0.10.3`

Nazeh:
I get a 404 error too but I don't get any indication on the place of the image, and the `html: { useGatsbyImage: false}` don't solve the issue, I would be happy if at least I can tell where the image is used!

```
info  gatsby-source-wordpress  Unrecoverable error
 occured while fetching media item #23787

Media item link: https://********.com/
?attachment_id=23787

 ERROR 

failed to process https://********.com/wp-content/uploads/2020/04/COVID-19-Event-Cancellations-1.png
HTTPError: Response code 404 (Not Found)
```

TylerBarnes:
Hi @ashhitch , this is actually by design to prevent deploying a broken site. We were considering this a feature because an admin can delete an image from the media library without knowing where that image lives. All pages or posts that had that image in html will then be broken. The link where the image is used is added to the error so that the site can be fixed https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/issues/20

@Nazeh your issue is a bit different actually. In your case you have a media item that exists in WordPress but the image file for it has been deleted on disk. I will look at improving that error message but you should be able to go to the media item with a db id of 23787 to see which media item is missing files.

TylerBarnes:
[...]

PR #208, opened 2020-10-09
Feat: add media item parent error context and allow 404s in develop

### LINKED ISSUE #206: Gatsby develop build fails when sourcing an image that returns a HTTP code 500
At the moment, if an image returns a 500, the whole build fails and prevents further development.

This kind of bug pretty much prevents building for production which would not be ideal for sites that are media heavy like the one I'm working on.

I think I would rather have a missing image on my site (that I am aware of because of a warning either in the command line or as a graphql error) than not be able to develop my website at all.

### Expected behavior
A warning similar to when an image returns an HTTP code 404, which currently looks like:
```zsh
warn  gatsby-source-wordpress

Received a 404 when trying to fetch {IMAGE URL}
from Post #12490 {POST TITLE}

Most likely this image was uploaded to this Post and then deleted from the media library.
You'll need to fix this and re-save this Post to remove this warning at
{POST EDIT URL}
```
This gives the developer an idea as to what images needs to be addressed without preventing them from developing further. 

### Actual Behavior
```zsh
ERROR 

gatsby-source-wordpress  failed to process {IMAGE_URL}
HTTPError: Response code 500 (Internal Server Error)

...

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! app-name@0.0.1 develop: `gatsby develop`
npm ERR! Exit status 1
```

TylerBarnes:
Closes #206 and #70

### New Features

- 404 images will no longer fail the build during `gatsby develop` but will continue to fail the build in production builds.
- Media item fetch errors now include the name of the parent step in which the MediaItem File node was being fetched.
- When fetching html media item files, non-404 error codes now return more helpful information about which media item is having the problem.
- 404ing images in production now have an extra line to the error message explaining that the build failed to prevent deploying a broken site.

### Changes

- MediaItem.remoteFile has been deprecated for a few months, querying for it now throws an error.

### DIFF
--- a/.test-runtime/src/components/template-parts/blog-post.js
+++ b/.test-runtime/src/components/template-parts/blog-post.js
@@ -16,9 +16,9 @@ function BlogPost({ data }) {
         {title}
       </Heading>
 
-      {!!featuredImage?.node?.remoteFile?.childImageSharp && (
+      {!!featuredImage?.node?.localFile?.childImageSharp && (
         <Box mb={5}>
-          <Img fluid={featuredImage.node.remoteFile.childImageSharp.fluid} />
+          <Img fluid={featuredImage.node.localFile.childImageSharp.fluid} />
         </Box>
       )}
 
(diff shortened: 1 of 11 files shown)


TylerBarnes:
Published in `gatsby-source-wordpress-experimental@1.7.0`

PR #210, opened 2020-10-09
Ignore fields that have no name

### LINKED ISSUE #169: WPGraphQL for ACF Causes Error on Gatsby Develop
When I install the WPGraphQL for ACF extension and follow the steps to add a ACF field group and the GraphQL Field Name for the ACF Group and then run `gatsby develop` I'm getting an error on build:

```
success onPreExtractQueries - 0.001s

ERROR 

Missing onError handler for invocation 'extracting-queries', error was 'Error: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "" does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "" does not.
```

The field group is showing up in GraphiQL explorer and am able to query it from outside of my WordPress installation such as:
image

But no luck within Gatsby.

Any ideas?

TylerBarnes:
Fixes #169

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## 1.7.1
+
+### Bug Fixes
+
+- Fields which had no name were causing errors. These fields are now silently excluded as they can't be used.
+
 ## 1.7.0
 
 ### New Features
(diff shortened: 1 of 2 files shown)


TylerBarnes:
Published in `gatsby-source-wordpress-experimental@1.7.1`

PR #213, opened 2020-10-09
Add structured reporting with codes to fetch-graphql

sslotsky:
Leverage the new reporter API whenever possible, so that error codes can be reported, allowing us to better categorize errors that originate from this plugin.

## What's inside?

This is a backwards compatible change that will fall back to the older reporter API for unsupported Gatsby versions. It consists of a pre-init step to set up the error map and a wrapper that relieves the caller of the responsibility of knowing whether or not the new API is supported.

- [x] Structure calls from `fetch-graphql`
- [ ] Structure the rest
- [ ] Testing strategy?
- [ ] Get buy-in on codes, levels, categories

### Error map

The error map is made up of codes representing the different kinds of errors that we might report from this plugin. An entry in the map must provide:

* The error level (e.g. warning, error)
* The error category (e.g. user, system, third party)
* A function to extract the error message from the report payload

The codes currently shown in this PR are only a proposal and are open for discussion.

### Wrapper

The wrapper function returns an object that provides a consistent API for using both the new reporter and the old.

* If an unsupported version of Gatsby is being used, the original `reporter` object will be returned from the wrapper, and additional arguments will be ignored.
* For newer versions of Gatsby, the wrapper accepts a message and a code that it will use to provide a structured payload.
* Consumers of the wrapper can always call `rep.panic(message, code)` and not worry about whether or not the new reporter API is supported.
* For newer versions of Gatsby, consumers of the wrapper can still omit the `code` argument, and `reporter` will fall back to the default identifier.


src/steps/index.js
```diff
@@ -1,10 +1,11 @@
...
-export { ensurePluginRequirementsAreMet } from "~/steps/check-plugin-requirements"
-export { ingestRemoteSchema } from "~/steps/ingest-remote-schema"
-export { persistPreviouslyCachedImages } from "~/steps/persist-cached-images"
-export { sourcePreviews } from "~/steps/source-nodes/update-nodes/source-previews"
-export { sourceNodes } from "~/steps/source-nodes"
-export { createSchemaCustomization } from "~/steps/create-schema-customization"
-export { setImageNodeIdCache } from "~/steps/set-image-node-id-cache"
-export { startPollingForContentUpdates } from "~/steps/source-nodes/update-nodes/content-update-interval"
-export { checkIfSchemaHasChanged } from "~/steps/ingest-remote-schema/diff-schemas"
+export { setGatsbyApiToState } from "~/steps/set-gatsby-api-to-state";
```
sslotsky, line 1: Sorry for the formatting changes :disappointed:

TylerBarnes: There's a .prettierrc file in the root :) if you run prettier using that it will fix it

sslotsky: Not sure why my VSCode isn't respecting that config

src/utils/report.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * First draft for error code categories and values.
+ * Very much up for discussion.
```
TylerBarnes, line 3: Looks like a good start :) might be good to add a note about how this relates to setErrorMap when adding new codes. Also might be good to keep CODES and setErrorMap in the same place since it looks like you need to add a CODE and manually map it in setErrorMap

TylerBarnes (changes requested):

TylerBarnes:
This looks awesome @sslotsky ! Thanks for doing this :)

src/utils/report.js
```diff
@@ -0,0 +1,42 @@
...
+        },
+      });
+    };
+  }
+
+  return {
+    panic: structuredReport("panic"),
+    error: structuredReport("error"),
+    warn: structuredReport("warn"),
+    log: structuredReport("log"),
```
julienp, line 40: I am not sure if `log` and `warn` should be wrapped,

src/steps/set-error-map.js
```diff
@@ -0,0 +1,43 @@
...
+        text: (context) => context.message,
+        level: "ERROR",
+        category: "USER",
+      },
+      [CODES.Timeout]: {
+        text: (context) => context.message,
+        level: "ERROR",
+        category: "USER",
+      },
+      [CODES.UnknownError]: {
```
julienp, line 36: Should we have this? For now we are treating errors without a code as "unknown". I would lean towards removing this and trying to have more precise errors.

However in core we have a example that kind of supports this usage. Error id `85901` seems to be a generic graphql error. It tells us that the error is graphql related, but not much more. Similarly this error could mean "unknown/generic error in Wordpress plugin", although I'd prefer to solve that by making sure we include the plugin name in structured logs.

sslotsky: I am happy to go either way on that one. If nobody has has a differing opinion I have no problem removing this one.

src/steps/index.js
sslotsky: @TylerBarnes I'm not finding the prettier config in this project. If there's actually not one I'd be glad to steal one from wherever, merge that separately, and then rebase this one.

src/steps/set-error-map.js
sslotsky: Speaking of 85901... is there any logic for how these codes are chosen? Guidelines for coming up with good codes? Trying to come up with a scheme that will allow me to create codes that are unique, meaningful, and easy to understand. Drawing a blank so far.

julienp: My understanding is that the first 3 digits are the group (arbitrarily chosen) and then we count inside that group. I read that somewhere, but I can't find it now. I think @LekoArts knows!

LekoArts: Here's an explanation (also see following message):

https://github.com/gatsbyjs/gatsby/pull/27176#issuecomment-703611482

julienp: Thank you, you're the best ❤️

sslotsky: Excellent, thank you for this. The ability to search for the code is one advantage I hadn't considered so I may rethink these codes a bit based on that.

sslotsky:
Superseded by #234

PR #229, opened 2020-10-13
Use new plugin options validation API

sslotsky:
## Why this PR?

This is part of our efforts to improve error handling in source plugins. It leverages a new standard API for validating plugin options using Joi.

## How to test this PR

I don't know yet. 

## What's left to do?

- [x] Get advice on testing


docs/plugin-options.md
```diff
@@ -167,6 +191,14 @@ When true, all internal GraphQL queries generated during node sourcing will be w
 },
 ```
 
+## debug.timeBuildSteps
+
+_TODO_
+
+## debug.disableCompatibilityCheck
```
TylerBarnes, line 198: @sslotsky on the call I mentioned taking these out, but I think let's leave them in like you have them as todo's. I'll make sure to mark some of them as internal options that could be removed at any time in the docs here. 🤔 maybe I should go through and prefix the options with ___internalOptionName in the code too or something.. This disableCompatibilityCheck one is one people shouldn't be using, but it's useful for testing the source plugin with new unsupported WPGraphQL versions. Technically anyone could find it in the source code and use it, but my thinking before was if it's undocumented, we can break it at any time 🤔

src/steps/declare-plugin-options-schema.js
```diff
@@ -0,0 +1,76 @@
+const queryTypes = [
+  `UserToMediaItemConnection`,
+  `WpContentNodeToEditLockConnectionEdge`,
+  `WPPageInfo`,
+  /* TODO: finish this list */
+]
+
+export function declarePluginOptionsSchema({ Joi }) {
```
TylerBarnes, line 8: We wont be able to get a predefined list of types at this point since the WP site owner registers the available types. We do have the plugin options though, so you should be able to get the keys they're currently using :

```
export function declarePluginOptionsSchema({ Joi }, pluginOptions) {
    const queryTypes = Object.keys(pluginOptions.type)
```

src/steps/declare-plugin-options-schema.js
```diff
@@ -0,0 +1,90 @@
...
+    excludeFieldNames: Joi.array().items(Joi.string()),
+    html: Joi.object({
+      useGatsbyImage: Joi.boolean().default(true),
+      imageMaxWidth: Joi.number().integer(),
+      fallbackImageMaxWidth: Joi.number().integer().default(100),
+      imageQuality: Joi.number().integer().default(90),
+      createStaticFiles: Joi.boolean().default(true),
+    }),
+    type: Joi.object().custom((value, helpers) => {
+      const reservedKeys = [`MediaItem`, `ContentNode`, `TermNode`, `Menu`]
```
TylerBarnes, line 48: @sslotsky I believe the only reserved key should be `MediaItem`, all other types take the same options. MediaItems also take the same options except for [`MediaItem.lazyNodes`](https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/docs/plugin-options.md#typemediaitemlazynodes-boolean) and [`MediaItem.localFile.excludeByMimeTypes`](https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/docs/plugin-options.md#typemediaitemlocalfileexcludebymimetypes-array) which the other types don't have.

TylerBarnes:
@sslotsky I think we can just bump the minimum peerdep version of Gatsby here 🤔 we don't really need to support older Gatsby versions. We can tell folks to update. Maybe it'd be nice to check the installed Gatsby version and throw an error saying to update 🤔

sslotsky:
> @sslotsky I think we can just bump the minimum peerdep version of Gatsby here  we don't really need to support older Gatsby versions. We can tell folks to update. Maybe it'd be nice to check the installed Gatsby version and throw an error saying to update

@TylerBarnes Sanity does something similar https://github.com/sanity-io/gatsby-source-sanity/blob/main/src/gatsby-node.ts#L61-L67

I'm looking through Gatsby's changelog to see what the right minimum version should be. Do you know a better way? Right now I've got [2.24.74](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/CHANGELOG.md#22474-2020-10-12) locally and that looks fine when I run tests.

TylerBarnes:
tbh it's hard to tell since the changelog there isn't very good usually. We know for sure that that version has the api at least, so lets go with that one 👍 to me it's relatively low friction for someone to update their Gatsby version. I'll make sure our official starter has the latest Gatsby version when I add it. This will be a breaking change version for this plugin which is ok

plugin/src/steps/declare-plugin-options-schema.js
```diff
@@ -0,0 +1,92 @@
...
+
+      Joi.assert(value.ContentNode, nodeSchema, `type.ContentNode`)
+      Joi.assert(value.TermNode, nodeSchema, `type.TermNode`)
+      Joi.assert(
+        value.Menu,
+        Joi.object({
+          beforeChangeNode: Joi.function().arity(1),
+        }),
+        `type.Menu`
+      )
```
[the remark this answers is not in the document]
sslotsky: Ok so if I understand you correctly, I think I end up with this?

```ts
      Joi.assert(
        value.MediaItem,
        Joi.object({
          exclude: Joi.boolean(),
          excludeFieldNames: Joi.array().items(Joi.string()),
          lazyNodes: Joi.boolean().default(false),
          localFile: Joi.object({
            excludeByMimeTypes: Joi.array().items(Joi.string()),
          }),
          beforeChangeNode: Joi.function().arity(1),
        }),
        `type.MediaItem`
      )

      dynamicKeys.forEach((k) => {
        Joi.assert(
          value[k],
          Joi.object({
            exclude: Joi.boolean(),
            excludeFieldNames: Joi.array().items(Joi.string()),
            nodeInterface: Joi.boolean(),
            beforeChangeNode: Joi.function().arity(1),
          })
        )
      })
```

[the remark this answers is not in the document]
sslotsky: Or maybe `MediaItem` should also have `nodeInterface`?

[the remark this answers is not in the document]
sslotsky: If that's the case, then how about this?

```js
      const baseSchema = Joi.object({
        nodeInterface: Joi.boolean(),
        exclude: Joi.boolean(),
        excludeFieldNames: Joi.array().items(Joi.string()),
        beforeChangeNode: Joi.function().arity(1),
      })

      Joi.assert(
        value.MediaItem,
        baseSchema.append({
          lazyNodes: Joi.boolean().default(false),
          localFile: Joi.object({
            excludeByMimeTypes: Joi.array().items(Joi.string()),
          }),
        }),
        `type.MediaItem`
      )

      dynamicKeys.forEach((k) => {
        Joi.assert(value[k], baseSchema, `type.${k}`)
      })
```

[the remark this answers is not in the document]
TylerBarnes: That looks great! 👍 none of these are required right? I haven't used Joi. I'd assume they're not unless you add something like `.isRequired()`, is that right?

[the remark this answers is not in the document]
TylerBarnes: To clarify, they should be optional :p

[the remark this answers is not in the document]
TylerBarnes: Oh 🤦 I just remembered that MediaItem's can't currently be excluded due to their special handling for gatsby-image. I'm not sure if that matters for this or not because the source plugin will error if you try to add MediaItem.exclude = true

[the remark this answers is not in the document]
sslotsky: I'll find a way to handle this here!

plugin/src/steps/declare-plugin-options-schema.js
```diff
@@ -0,0 +1,77 @@
...
+        exclude: Joi.boolean(),
+        excludeFieldNames: Joi.array().items(Joi.string()),
+        beforeChangeNode: Joi.function().arity(1),
+      })
+
+      Joi.assert(
+        value.MediaItem,
+        baseSchema.append({
+          lazyNodes: Joi.boolean().default(false),
+          exclude: Joi.boolean().valid(false),
```
sslotsky, line 64: @TylerBarnes by adding this to the `append` statement we prevent them from specifying `type.MediaItem.exclude = true`

duffn:
@TylerBarnes Is this good to merge?

TylerBarnes:
@duffn it's close but not quite. This will be a breaking change so I'm waiting to merge this until the new preview re-work is also merged (since that will be a breaking change too) so that folks don't need to do 2 major updates within the span of a week or so. I also need to pull it down again and do some manual testing against various plugin configurations. There's some likely dead code now related to plugin option validation in https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/plugin/src/steps/process-and-validate-plugin-options.js that I'd like to check through and delete. And the other thing is we need to rebase or merge master again because this repo has been converted to a monorepo since this PR was opened 🙈

TylerBarnes:
Oh we also might need to detect the Gatsby version and throw an error saying to update gatsby core depending on what the DX is like for folks who update but are on an old Gatsby version 🙈

sslotsky:
> Oh we also might need to detect the Gatsby version and throw an error saying to update gatsby core depending on what the DX is like for folks who update but are on an old Gatsby version

@TylerBarnes I did look into that actually, [more details in this Slack thread](https://gatsbyjs.slack.com/archives/CA1GW1HNU/p1603912091313000)

TylerBarnes:
Thanks @sslotsky ! So to summarize, Gatsby will show a warning and (now that pluginOptionsSchema should be released), it will provide some additional info? We'll need to bump the peerDep version of Gatsby then and check what the warning looks like. I've seen Gatsby sites before where there are 50 ignored warnings on every build 🤦

sslotsky:
> We'll need to bump the peerDep version of Gatsby then

@TylerBarnes yes thank you, the `package.json` moved on me so I missed it. Just pushed.

And yeah I agree, just having warnings here is really far from ideal. Wish we could fail the build, but it seems plausible that some users will ignore the warnings and then the build will fail further down the line, and the reason for the failure will be much less clear.

TylerBarnes:
For sure. I think we can fail the build tho. We're already using the semver package inside this plugin to create some prettier errors around remote plugin versions. So we could use that same package to check if the installed Gatsby version in package.json is after our required version and panic if it's not

plugin/src/steps/check-gatsby-version.js
```diff
@@ -0,0 +1,23 @@
+const semver = require(`semver`)
+const fs = require(`fs`)
+
+export function checkGatsbyVersion({ reporter }) {
```
sslotsky, line 4: @TylerBarnes I think this is how we'd do the version check we discussed, yeah?

TylerBarnes: This looks good to me 👍 I think you mean minimum instead of minimim though :)

sslotsky: LOL

sslotsky: Let me know if you want any other tweaks to that message

TylerBarnes: It might be helpful for users if we mention why it's a requirement. Something like `The minimum supported Gatsby version is 2.24.74. This is because previous versions do not support the new "pluginOptionsSchema" node API which this plugin is now implementing.` or something like that

TylerBarnes:
Since #268 was a duplicate of this PR and both this PR and that PR had parts missing from each other, I've manually pulled in @mxstbr 's changes here. After doing that I've added some missing options and allowed `null` and `false` for many of the options since I've seen folks making these null and false depending on the existence of an environment var.

Copying @mxstbr 's PR description here:

```
This will cause Gatsby to validate the options users pass to this plugin! 🎉 See https://www.gatsbyjs.com/docs/configuring-usage-with-plugin-options/#how-to-validate-plugin-options for more information.

We need this in order to add WordPress support to the new create-gatsby command. This will cause create-gatsby to prompt users for the required options (i.e. in this case only the url) and help them set the plugin up correctly: gatsbyjs/gatsby#27703

I initially added all the options from the readme, however when I tried to validate the options passed to the test-site, there were a couple in there that aren't documented.

I added the ones from the test-site as well, but of course the question is: are there any more undocumented options that people use? 🤔
```

Upon doing some manual testing on what the experience was like for folks on old Gatsby versions I've changed course and removed the semver check and build fail when Gatsby doesn't support the `pluginOptionsSchema` api. It's much nicer to make this non-breaking and allow folks to upgrade at their leisure. (Sorry for that @sslotsky 😅 ) I realized that folks who might be on latest Gatsby might still have a version range in their `package.json` which doesn't necessarily satisfy the version of Gatsby that's required for this api. So this could be a more annoying breaking change than I initially thought.

Thanks so much for this @sslotsky and @mxstbr ! Apologies on the delay on getting this merged.

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@2.2.0`. Thanks again!!

PR #230, opened 2020-10-13
add previously gitignored files 🤦‍♂️

TylerBarnes:
So since I've been working on this repo alone for the last year, I didn't even realize I just had these files locally haha..


TylerBarnes:
The config files here were copied from the Gatsby core repo to keep things consistent when we merge this back in

PR #234, opened 2020-10-14
Add codes to reporter.panic in fetch-graphql

sslotsky:
Supersedes #213 

# Adding codes to reporter.panic

Leverage the new reporter API whenever possible, so that error codes can be reported, allowing us to better categorize errors that originate from this plugin.

## What's inside?

This is a backwards compatible change that will fall back to the older reporter API for unsupported Gatsby versions. It consists of a pre-init step to set up the error map and a wrapper that relieves the caller of the responsibility of knowing whether or not the new API is supported.

### Error map

The error map is made up of codes representing the different kinds of errors that we might report from this plugin. An entry in the map must provide:

* The error level (e.g. warning, error)
* The error category (e.g. user, system, third party)
* A function to extract the error message from the report payload

The codes currently shown in this PR are only a proposal and are open for discussion.

### Wrapper

The wrapper function returns an object that provides a consistent API for using both the new reporter and the old.

* If an unsupported version of Gatsby is being used, the original `reporter` object will be returned from the wrapper, and additional arguments will be ignored.
* For newer versions of Gatsby, the wrapper accepts a message and a code that it will use to provide a structured payload.
* Consumers of the wrapper can always call `rep.panic(message, code)` and not worry about whether or not the new reporter API is supported.
* For newer versions of Gatsby, consumers of the wrapper can still omit the `code` argument, and `reporter` will fall back to the default identifier.


src/utils/fetch-graphql.js
```diff
@@ -80,13 +86,19 @@ const handleGraphQLErrors = async ({
...
 
   if (!response) {
-    reporter.panic(response)
+    /**
+     * FIXME
+     *
+     * 1. We're logging an undefined variable here instead of a message.
+     * 2. What scenarios lead us to this point? Can we assign a code?
+     */
+    structuredReporter.panic(response)
```
sslotsky, line 101: cc @TylerBarnes any thoughts here?

TylerBarnes: Uhh 🤔 wth hahaha. This is definitely an accident.. This is what happens when you don't have code review 🤦 .

TylerBarnes: I'll fix this separately from this PR. Feel free to leave your FIXME in there

TylerBarnes: The fact that there's also a return after it panics is a big LOL

src/utils/fetch-graphql.js
```diff
@@ -443,9 +470,11 @@ const fetchGraphql = async ({
       throw e
     }
 
+    const structuredReporter = makeStructuredReporter(reporter)
```
TylerBarnes, line 473: structuredReporter is being created conditionally, but lower down on L504, it's called but is undefined.

I wonder if it would be good to make a reporter helper that gets the reporter, wraps it in structuredReporter, and adds formatting to the input error message 🤔 

Maybe that's outside the scope of this PR, but since all errors need to be wrapped in this PR it might be the best time to do it. It'll make your logic easier too since you could just do:

```
import { wpReporter } from "~/utils/wp-reporter"

wpReporter.error(`Yo this is broken!`)
```

TylerBarnes: @sslotsky , I've added a helper and set the reporter to it using the public Gatsby api instead of directly importing from gatsby-cli

sslotsky:
@TylerBarnes I think our branch sharing shenanigans put me in a weird state so I decided to:

1. Reset back to the commit before we started sharing
1. Add some commits to get things working correctly
1. Force push the branch

So I apologize for blowing away your helper changes. Now that things are working against the newer reporter API, the next thing I'd want to prioritize is making sure that it's still working for the old version. If we want to reintroduce your helper I'd be okay with doing it after that or in a future PR.

sslotsky:
Lennart outlined a process for testing against the old version [in this Slack thread](https://gatsbyjs.slack.com/archives/C01AJ60GRUK/p1602752023054100), I'll look into that this morning!

src/utils/fetch-graphql.js
```diff
@@ -15,14 +16,14 @@ const handleErrorOptions = async ({
...
-  reporter,
+  structuredReporter,
 }) => {
   if (
     variables &&
     Object.keys(variables).length &&
     pluginOptions.debug.graphql.showQueryVarsOnError
   ) {
-    reporter.error(
+    structuredReporter.error(
```
TylerBarnes, line 26: @sslotsky maybe we should have a debug code. In this case .error is being called but this is supplemental info to other errors that are occurring. It'd be good to be able to mark these as debugging errors maybe 🤔  what do you think?

src/utils/fetch-graphql.js
```diff
@@ -114,9 +126,15 @@ const handleGraphQLErrors = async ({
       error.message === errorMap.from
 
     if (errorWasMapped && panicOnError) {
-      reporter.panic(formatLogMessage(errorMap.to))
+      structuredReporter.panic(
+        formatLogMessage(errorMap.to),
+        CODES.GQLConfiguration
```
TylerBarnes, line 131: Rathe than configuration, these are actual GraphQL errors being returned from the remote schema. Maybe we need another code for remoteGraphQLError?

src/utils/fetch-graphql.js
sslotsky: Sure, I can add a debug code!

src/utils/fetch-graphql.js
```diff
@@ -260,33 +280,36 @@ const handleFetchErrors = async ({
...
   const forbidden = e.message.includes(`Request failed with status code 403`)
 
   if (forbidden) {
-    reporter.panic(
+    structuredReporter.panic(
       formatLogMessage(
         `${e.message}\n\nThe GraphQL request was forbidden.\nIf you are using a security plugin like WordFence or a server firewall you may need to whitelist your IP address or adjust your firewall settings for your GraphQL endpoint.\n\n${errorContext}`
-      )
+      ),
+      CODES.Permissions
```
TylerBarnes, line 295: I think this code might be confusing in the context of GraphQL errors since GraphQL requests can have data level permissions. I wonder if this would make more sense as `RequestPermissions` or `RequestDenied`

src/utils/fetch-graphql.js
```diff
@@ -260,33 +280,36 @@ const handleFetchErrors = async ({
...
       formatLogMessage(
         `${e.message}\n\n${errorContext}\n\nThis can happen due to custom code or redirection plugins which redirect the request when a post is accessed.\nThis redirection code will need to be patched to not run during GraphQL requests.\n\nThat can be achieved by adding something like the following to your WP PHP code:\n
 if ( defined( 'GRAPHQL_REQUEST' ) && true === GRAPHQL_REQUEST ) {
   return examplePreventRedirect();
 }
 
 ${slackChannelSupportMessage}`
-      )
+      ),
+      CODES.CustomUserCode
```
TylerBarnes, line 312: This should be WordPressFilters I think. The custom code that causes redirections is on the WP side and uses filters

src/utils/fetch-graphql.js
```diff
@@ -329,17 +352,19 @@ ${slackChannelSupportMessage}`
         {
           useVerboseStyle: true,
         }
-      )
+      ),
+      CODES.BadUrl
```
TylerBarnes, line 356: In this case I think BadResponse makes more sense. The url might be right but there could be a problem with the WP instance. So this could be BadResponse or HtmlResponse

TylerBarnes (changes requested):

TylerBarnes:
I'm getting `Cannot read property 'toString' of undefined` when starting Gatsby develop

Screen Shot 2020-10-15 at 11 41 57 AM

TylerBarnes:
Ahh you know what, I added a new error to test but forgot to include an error code 😆 

Maybe we should catch that and throw a new error explaining to add an error code? I guess then we might need a CODE.noErrorCode 🤣

TylerBarnes:
I'm getting that joi error again:

Screen Shot 2020-10-15 at 1 56 21 PM

```
Failed to validate error Error [ValidationError]: child "code" fails because ["code" must be a string]
    at Object.exports.process (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/node_modules/@hapi/joi/lib/errors.js:202:19)
    at internals.Object._validateWithOptions (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/node_modules/@hapi/joi/lib/types/any/index.js:763:31)
    at internals.Object.validate (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/node_modules/@hapi/joi/lib/types/any/index.js:797:21)
    at constructError (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/node_modules/gatsby-cli/lib/structured-errors/construct-error.js:52:32)
    at Reporter.error (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/node_modules/gatsby-cli/lib/reporter/reporter.js:135:59)
    at Object.panic (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/node_modules/gatsby-cli/lib/reporter/reporter.js:75:34)
    at handleGraphQLErrors (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/gatsby-source-wordpress-experimental/src/utils/fetch-graphql.js:131:16)
    at fetchGraphql (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/gatsby-source-wordpress-experimental/src/utils/fetch-graphql.js:521:11)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Promise.all (index 0)
    at ensurePluginRequirementsAreMet (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/gatsby-source-wordpress-experimental/src/steps/check-plugin-requirements.js:285:3)
    at runSteps (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/gatsby-source-wordpress-experimental/src/utils/run-steps.js:25:9)
    at runAPI (/Users/tyler/Documents/GitHub/gatsby-wordpress-workspace/node_modules/gatsby/src/utils/api-runner-node.js:350:16) {
  isJoi: true,
  details: [
    {
      message: '"code" must be a string',
      path: [Array],
      type: 'string.base',
      context: [Object]
    }
  ],
  _object: {
    context: {
      sourceMessage: '\u001b[44m\u001b[37m gatsby-source-wordpress \u001b[39m\u001b[49m WPGatsby is not active in your WordPress installation.\n' +
        'To download the latest versions of WPGatsby and WPGraphL, visit:\n' +
        'https://github.com/TylerBarnes/using-gatsby-source-wordpress-experimental/tree/master/WordPress/plugins'
    },
    text: '\u001b[44m\u001b[37m gatsby-source-wordpress \u001b[39m\u001b[49m WPGatsby is not active in your WordPress installation.\n' +
      'To download the latest versions of WPGatsby and WPGraphL, visit:\n' +
      'https://github.com/TylerBarnes/using-gatsby-source-wordpress-experimental/tree/master/WordPress/plugins',
    level: 'ERROR',
    category: 'THIRD_PARTY',
    stack: [],
    docsUrl: 'https://gatsby.dev/issue-how-to',
    code: 112001
  },
  annotate: [Function]
}
```

TylerBarnes:
To reproduce this, I disabled WPGatsby on a local test WP instance. I can disable it on the live test instance if that'd help!

TylerBarnes:
Ok, I disabled it so you should see the error when you run `gatsby develop` in `.test-runtime`

TylerBarnes (approved):
Thanks Sam! This is really great. Much appreciated :)

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@1.7.7`. Thanks again!

PR #240, opened 2020-10-15
setup to run tests in docker

acao:
This is pretty close to working

just try `docker-compose up -d` to try it out.

when i run yarn test, i see this:

```
Client side block registry is missing. You need to open up gutenberg or load it from WPGraphQLGutenberg Admin page.
```

so now i just need figure out how to do this programatically?


docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+    image: mysql:8
+    environment:
+      MYSQL_ROOT_PASSWORD: gtsb-wp-dckr
+      MYSQL_DATABASE: wordpress
+      MYSQL_USER: wordpress
+      MYSQL_PASSWORD: gtsb-wp-dckr-user
+    restart: always
+    volumes:
+      - "db_data:/var/lib/mysql"
+      - "./docker/seed:/docker-entrypoint-initdb.d"
```
acao, line 14: @TylerBarnes this is where we import the seed file

docker/wp-cli/start.sh
```diff
@@ -0,0 +1,34 @@
...
+    --title="Gatsby & WordPress" \
+    --admin_user=admin \
+    --admin_password=secret \
+    --admin_email=admin@admin.com
+
+wp search-replace 'https://gatsbyinttests.wpengine.com' 'http://localhost:8081'
+
+wp user update admin --user_pass="secret"
+# install plugins that the wpgraphql plugins build on
+wp plugin install gutenberg advanced-custom-fields wordpress-seo
```
acao, line 19: @TylerBarnes we need to pin to specific (provided?) plugin versions here. so we should start with the versions running on the instance where this seed file comes from!

My theory is that this is will resolve the error I'm still seeing wordpress side when querying the schema in the wp-graphql graphiql:

```
PHP Notice:  Undefined index: ID in /var/www/html/wp-content/plugins/advanced-custom-fields/includes/acf-field-functions.php on line 361, referer: http://localhost:8001/wp-admin/admin.php?page=graphiql-ide

```

docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+    # image: duffn/wordpress-wpgraphql-wpgatsby:latest
+    # While pulling the image will be faster, if you like, you can build the
+    # image locally, but uncommenting the lines below and commenting out the
+    # `image` line above.
+    # The image above will be kept up-to-date, however, you can change WPGraphQL
+    # and WPGatsby versions below if you build yourself.
+    build:
+      context: ./docker/wordpress
+      args:
+        WPGRAPHQL_VERSION: v0.13.3
```
acao, line 28: this is where we have to pin the prebuild docker image plugin versions (more ideal for speed)

build args don't accept `env_file` yet unfortunately, but maybe there's a way around that issue i'm overlooking? @jasonbahl i saw you using template replacement in the wp-graphql docker compose, how is that compiled? is that just a new docker-compose feature i'm unaware of?

however, the wp-cli image runs everything when the container runs, so we can use an env file to specify those plugin versions

docker/wordpress/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+  && mv composer.phar /usr/local/bin/composer \
+  && curl -O 'https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar' \
+  && chmod +x wp-cli.phar \
+  && mv wp-cli.phar /usr/local/bin/wp
+
+
+COPY install-plugin.sh /usr/local/bin/install-plugin
+
+# WP GraphQL
+RUN install-plugin wp-graphql wp-graphql/wp-graphql ${WPGRAPHQL_VERSION}
```
acao, line 28: @jasonbahl and then we install the plugins here, using just the release tarball for now. I tried getting composer to work here in the image build, but it didn't pan out because of user permissions issues I couldn't reconcile. thinking maybe we can run composer in the wp-cli process, where the user permissions are set up better for installing plugins?

TylerBarnes: @acao you might be able to install a plugin outside of wp-cli and then activate it with wp-cli

TylerBarnes:
@acao I'm getting this:

```
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

TylerBarnes:
my bad, haha! I just had to update docker desktop 😜

TylerBarnes:
It looks like the tests are failing because the docker WP instance is missing the image files. I'm totally lost when it comes to docker, how do we get those added?

acao:
added! though i think i might try to overwrite with a zip because now the git archive is YUGE. i think about half the media files or more are not needed

acao:
made a ton of progress on this, but i will need to split it into two efforts:

1. first, a PR to move the repo to a monorepo
2. secondly, support running tests in docker, add CI, and update snapshots

PR #241, opened 2020-10-16
Fix/build setup

TylerBarnes:
This PR allows us to write typescript & makes it easier to setup node debugging. JS/TS is compiled to `.compiled` now with source maps when running `yarn watch` and without for `yarn build`


acao:
looks great! `.compiled` is a bit awkward, can we just use `dist`? i wonder if the default eslint/babel/etc configs expect the standard typescript output directory?

TylerBarnes:
Sure thing :) one sec

TylerBarnes:
Updated it!

PR #248, opened 2020-10-24
chore: docker tests

acao:
And now we have tests running in docker, and in CI! :D


acao:
@TylerBarnes in case you're curious, i think the issue here is that in the jest teardown, `tree-kill` when executed does not have sufficient permission. this seems to very specifically be a permissions error with `kill`. i wonder if there's another way we can do this?
```
 errno: 'EPERM',
 code: 'EPERM',
 syscall: 'kill'
```

acao:
got it! now for some cleanup

TylerBarnes:
Woo! 😄 awesome!

acao:
image

having this issue, so i'm trying to bump to the approproate wp-graphql version. probably should just use composer but this gives us at least 0.14.0

acao:
so now the build suites pass, but the schema one doesn't, after bumping wp-graphql. so this is as expected, correct @jasonbahl  and @TylerBarnes ?

TylerBarnes:
@acao I'll pull it down and have a look :)

docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+    # `image` line above.
+    # The image above will be kept up-to-date, however, you can change WPGraphQL
+    # and WPGatsby versions below if you build yourself.
+    build:
+      context: ./docker/wordpress
+      args:
+        # specify version updates here
+        # this is for plugins installed via git or tar
+        WPGRAPHQL_VERSION: v0.14.0
+        WPGATSBY_VERSION: v0.4.17
```
TylerBarnes, line 31: Latest WPGatsby is 0.5.4 :)

TylerBarnes (changes requested):
This is so amazing, thank you @acao !!

Just left a note about the WPGatsby version.
Beyond that I'm having a hard time getting my docker setup rebuilding to use the newly specified version of WPGraphQL. I ran `yarn docker-start --build --force-recreate` and then tried to fully deleted the container and ran yarn docker-start but in both cases it says the WP instance has the wrong versions of WPGraphQL and WPGatsby.

If possible it'd be great to have a section in the docs explaining how to use the docker tests and what steps are required to test different versions of WPGraphQL and WPGatsby.

Thanks again :)

docker-compose.yml
acao: got it!

test-site/gatsby-config.js
```diff
@@ -13,82 +13,7 @@ require(`dotenv`).config({
 
 // this is it's own conditional object so we can run
 // an int test with all default plugin options
-const wpPluginOptions = !process.env.DEFAULT_PLUGIN_OPTIONS
```
acao, line 16: @TylerBarnes is this what we want to do? so no need for the env var anymore if we're only running it once, right? or is it that we want the former for the schema int suite, and the empty object (formerly w/ cart b/c  woocommerce) for the build int suite?

acao:
oh 5min at most, not bad!

test-site/__tests__/integration/schema/data-resolution.test.js
```diff
@@ -52,96 +51,96 @@ describe(`[gatsby-source-wordpress-experimental] data resolution`, () => {
...
-      JSON.stringify(WPGraphQLData.page).replace(
-        /https:\/\/gatsbyinttests.wpengine.com/gm,
-        ``
-      )
-    )
-
-    expect(gatsbyResult.data.wpPage).toStrictEqual(wpGraphQLPageNormalizedPaths)
-    expect(gatsbyResult.data.wp.seo).toStrictEqual(WPGraphQLData.seo)
-  })
+  // testResolvedData({
```
acao, line 54: @TylerBarnes delete or todos? i guess we can address that in another PR?

docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+      MYSQL_PASSWORD: gtsb-wp-dckr-user
+    restart: always
+    volumes:
+      - "db_data:/var/lib/mysql"
+      - "./docker/seed:/docker-entrypoint-initdb.d"
+
+  wordpress:
+    depends_on:
+      - db
+    # image: duffn/wordpress-wpgraphql-wpgatsby:latest
```
duffn, line 19: I think you could drop these comments.

docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+  wordpress-cli:
+    depends_on:
+      - db
+      - wordpress
+    # Specify versions for wp cli or composer here
+    environment:
+      CPT_UI_VERSION: 1.8.1
+    env_file:
+      - ./test-site/.env.test
+    # image: duffn/wp-cli-gatsby:latest
```
duffn, line 57: Same here, don't really need all these comments.

docker/wordpress/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+ARG WPGRAPHQL_VERSION=v0.13.1
+ARG WPGATSBY_VERSION=v0.5.2
+ARG WPGRAPHQL_ACF_VERSION=v0.3.1
+ARG WPGRAPHQL_CPT_UI_VERSION=v1.1
+ARG ACF_VERSION=v5.8.7
+ARG WP_CONTENT_DIR=/usr/src/wordpress/wp-content
+ARG PLUGIN_DIR=${WP_CONTENT_DIR}/plugins
+ARG UPLOADS_DIR=${WP_CONTENT_DIR}/uploads
+
+LABEL maintainer="duffn@outlook.com"
```
duffn, line 12: Do you want to make this you?

docker/wp-cli/start.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
+
+echo "heyyyy"
```
duffn, line 4: I guess this was from debugging.

docker/wp-cli/start.sh
```diff
@@ -0,0 +1,29 @@
...
+
+echo "heyyyy"
```
duffn, line 6: Heyyyyyyy. Remove me.

test-site/__tests__/integration/schema/data-resolution.test.js
duffn: I was going to comment the same. Delete commented out bits to keep it clean.

duffn (commented):
Looks good. Only minor nits about cleaning up some comments and code left over from debugging!

acao:
@duffn got it! yes good catch, another cleanup commit on the way

test-site/gatsby-config.js
```diff
@@ -36,25 +36,23 @@ const wpPluginOptions = !process.env.DEFAULT_PLUGIN_OPTIONS
...
-          where: `language: FR`,
-          beforeChangeNode: async ({ remoteNode, type }) => {
-            if (remoteNode.language && remoteNode.language.slug !== `fr`) {
-              return {
-                cancelUpdate: true,
-              }
-            }
-          },
-        },
+        // TODO: test translation support using our own test wp graphql plugin
```
TylerBarnes, line 45: This was actually to test our where and beforeChangeNode type options :)

package.json
```diff
@@ -17,12 +17,17 @@
...
+    "ci": "yarn lint && yarn ci-test",
+    "docker-start": "docker-compose up",
+    "test-schema-first": "START_SERVER=true jest ./test-site/__tests__/integration/schema",
+    "test-schema-increment": "WPGQL_INCREMENT=true START_SERVER=true jest ./test-site/__tests__/integration/schema",
+    "test-schema": "yarn clean-test-runtime && yarn test-schema-first && yarn test-schema-increment",
+    "test-build": "DEFAULT_PLUGIN_OPTIONS=true jest ./test-site/__tests__/integration/build",
+    "test": "yarn build && yarn test-schema && yarn test-build && yarn clean-test-runtime",
+    "test-schema-watch": "SHOW_GATSBY_PROCESS_STDOUT=true yarn test-schema-first --watch",
+    "test-build-watch": "SHOW_GATSBY_PROCESS_STDOUT=true yarn test-build --watch",
+    "test-update": "yarn test-schema-first -u && yarn test-schema-increment -u && yarn test-build -u",
```
TylerBarnes, line 30: I think we want to remove the `-u` flags here

TylerBarnes (changes requested):
It'd be great if you could add a couple lines about how to run the tests and what steps are needed when the WPGraphQL or WPGatsby versions change here https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/docs/community-and-support.md#contributing

I'm not sure what I need to do to run the tests because my docker instance is stuck on the wrong WPGraphQL/WPgatsby versions 😅

package.json
acao: @TylerBarnes they are there on purpose!

test-site/gatsby-config.js
acao: ohh i see, should i just delete those fields then?

acao:
@TylerBarnes sure, i can leave some instructions on how to run it and clear it. these steps should allow you to clear and reset everything:

1. `docker-compose rm -fsv`
2. `docker volume prune` (you should see two related volumes referenced, say yes)
3. `docker-compose up -d`

then wait about 15-30s for the schema to be present for tests

TylerBarnes:
Thanks @acao ! It's running now which is great. I'm running into a timeout error which is odd because the timeout is set to 100 seconds but Jest seems to have it set to 10 seconds.

```
: Timeout - Async callback was not invoked within the 10000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 10000ms timeout specified by jest.setTimeout.Error:
```

```
jest.setTimeout(100000)
```

TylerBarnes:
Ahh, found it 🤦 looks like I set it to 10 seconds in `test-utils/jest.setup.js` about a month ago 😆

duffn:
> @TylerBarnes sure, i can leave some instructions on how to run it and clear it. these steps should allow you to clear and reset everything:
> 
>     1. `docker-compose rm -fsv`
> 
>     2. `docker volume prune` (you should see two related volumes referenced, say yes)
> 
>     3. `docker-compose up -d`
> 
> 
> then wait about 15-30s for the schema to be present for tests

I think we should suggest only deleting the volumes for this setup here and not pruning all local volumes.

```
docker volume rm app-name_whatever_this_volume_is
```

package.json
TylerBarnes: Ahhh 😆 oops, I didn't read the script name. That's useful!

test-site/gatsby-config.js
TylerBarnes: Maybe update the todo message so we remember to test that later 👍

TylerBarnes:
With that in mind ☝️ maybe it'd be good to add a yarn script like `docker-reinstall` or something.

acao:
@TylerBarnes added some basic contributing docs, and i made it so that the `yarn docker-start` command will automatically flush volumes when it's complete, which seems appropriate for tests. this way, we would never see it build without the declared versions in any case, unless we start the cluster in a different way

TylerBarnes:
Sweet, thanks @acao ! I noticed the default options build didn't have a `gatsby clean` before which means it wouldn't test the full process (a cached build runs about 1/20th of the code paths of the uncached build), so I added that quick. 

LGTM!

TylerBarnes (approved):

PR #255, opened 2020-10-29
chore: when checking coverage, preprocessor fails on transforming typescript

acao:
when running jest with `--coverage` in the repo (after configuring it to collect coverage from the plugin in this PR), i see this:
image

this is taken from the gatsby monorepo, and ensures we can transform typescript properly when parsing it, though we don't have suites we can collect coverage from yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ node_modules
 public
 .cache
 plugin/README.md
+coverage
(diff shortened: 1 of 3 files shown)


TylerBarnes (changes requested):
@acao on my preview branch I saw this as well, upgrading to the latest version of the Gatsby babel preset fixed it. Do you know if this will still be needed after upgrading that package?

TylerBarnes:
I'm 99% sure this is fixed but we can re-visit if we run into this again 👍

PR #256, opened 2020-10-29
Pedantic cleanup of inconsistent indentation

mor10:
Tabs and spaces should never mix


TylerBarnes:
Thanks @mor10 ! :D I hadn't noticed eslint wasn't working properly for the last while. Good news is we have a PR coming in that fixes and prevents these things. Thanks for this though!

PR #261, opened 2020-10-30
fix: for now, error when using the plugin multiple times. fixes #251

### LINKED ISSUE #251: Chore: temporarily disallow multiple instances of source plugin
For now adding more than one instance of the source plugin should throw an error. In the future we should allow multiple instances. Follow #58 for that

acao:
and toggle it as early as possible to prevent either plugin from executing!

i'd like to use `store.getState()` but it's not stable


TylerBarnes:
@acao I made a couple quick changes. The error was running when only 1 instance of the source plugin was installed, so I moved all the logic into the 1 file and have it flipping the var it's checking from false to true the first time. I updated the error message too since format log message already includes the plugin name.
It looks good to me but let me know what you think!

acao:
this is what i had implemented before, but it doesn’t  throw an error for either case :/

TylerBarnes:
@acao our code must've been slightly different since it's working

With 2 instances:

Screen Shot 2020-10-30 at 5 55 55 PM

With 1 instance:

Screen Shot 2020-10-30 at 5 56 42 PM

TylerBarnes:
Hmmm I'm just thinking about this and this will probably throw an error when createSchemaCustomization runs again (if there's 1 instance). I think this should probably go in onPreInit 🤔

acao:
ok, I'll look into that some more today!

acao:
@TylerBarnes oh i remember this from another PR. there is no `onPreBuild`! that explains why the other suggesion isn't working at all. and then `onPreBootstrap` almost works, but only happens once as well, same as with `createSchemaCustomization`. 

it seems `runApiInSteps` assumes these API hooks are run multiple times, but it seems it only fires once when i console log it. 

personally, i think we should just use gatsby store because this is temporary anyways, and the way we are using the store likely won't change. the flattenedPlugins array has been there for a while and each collection item has had `name` for a while.

acao:
@TylerBarnes works actually now!

my issue was that i had two instances of the same plugin config. your suggestion of `onPreInit` worked once I added a delta (appending a fake sub-path to the wp graphql URL worked for me)

also for some reason when i did `git pull origin --rebase` it preferred my additional commit I hadn't pushed, thus you'll see a weird diff, but everything should be here now!

I started writing a test but felt like dynamically generating the gatsby config was not going to be an easy path to go down. and unit tests wouldn't have the same purpose, as this case can really only be asserted via integration with gatsby CLI/core

TylerBarnes:
That makes sense! My guess is core makes a plugin id using `createContentDigest()` out of the plugin name + options. That's ok because if someone adds 2 identically configured instances of the source plugin, the expected behaviour would be that they would overwrite each other (imo) so this is good.
We should be ok without tests for this since it's a pretty minor thing anyway, when we do our documentation push, we'll have to make sure to document that only 1 instance works for now. I don't expect that we'll touch this code again until we allow multiple instances anyway.

Thanks again @acao !

TylerBarnes:
Oh also! `runApiSteps` isn't doing anything special. It runs all the "steps" in array order for the node api of the property name.  While doing that it adds an error boundary for each "step" and automatically includes the step name in any user facing errors that are thrown in that step. If there's an array within the array, all of those "steps" run in parallel with Promise.all.
I need to take 30 mins and sit down and add doc blocks throughout the codebase. I didn't originally add them because the codebase was in such a wild state of flux with massive pieces of code being deleted and replaced for a long time. That + no typescript were the biggest mistakes of this project :p

acao:
yeah, I can relate, haha!

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@2.1.0`

PR #262, opened 2020-11-02
fix: ensure that urls with query params are rewritten

rburgst:
- in case an anchor link has query params
  the URL was not properly rewritten


TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@3.1.2` thanks @rburgst !!

PR #268, opened 2020-11-04
Add pluginOptionsSchema

mxstbr:
This will cause Gatsby to validate the options users pass to this plugin! :tada: See https://www.gatsbyjs.com/docs/configuring-usage-with-plugin-options/#how-to-validate-plugin-options for more information.

We need this in order to add WordPress support to the new create-gatsby command. This will cause create-gatsby to prompt users for the required options (i.e. in this case only the `url`) and help them set the plugin up correctly: https://github.com/gatsbyjs/gatsby/pull/27703

---

I initially added all the options from the readme, however when I tried to validate the options passed to the `test-site`, there were a couple in there that aren't documented.

I added the ones from the `test-site` as well, but of course the question is: are there any more undocumented options that people use? 🤔


plugin/gatsby-node.js
```diff
@@ -27,3 +27,171 @@ module.exports = runApisInSteps({
...
+          excludeByMimeTypes: Joi.array()
+            .items(`string`)
+            .default([])
+            .description(
+              `Allows preventing the download of files associated with MediaItem nodes by their mime types.`
+            ),
+        }),
+      }),
+    })
+      .pattern(Joi.string(), getTypeOptions())
```
sslotsky, line 194: Aha! This must be the right way to handle what I've done here!

https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/pull/229/files#diff-7a6e1ab692943ea7446ae0997cceb01fa9ea1f60e3482b0398d80ce4b46d8659R72

package.json
```diff
@@ -24,7 +24,8 @@
     "test-schema-increment": "WPGQL_INCREMENT=true START_SERVER=true jest ./test-site/__tests__/integration/schema",
     "test-schema": "yarn clean-test-runtime && yarn test-schema-first && yarn test-schema-increment",
     "test-build": "yarn clean-test-runtime && DEFAULT_PLUGIN_OPTIONS=true jest ./test-site/__tests__/integration/build",
-    "test": "yarn build && yarn test-schema && yarn test-build && yarn clean-test-runtime",
+    "test-js": "yarn jest plugin/src",
```
acao, line 27: there are no tests in `plugin/src`, they are all in `test-site` workspace

acao:
@TylerBarnes should we mark this as a replacement PR for #229?

TylerBarnes:
Thanks for this @mxstbr and sorry for the delay in seeing this. My GH notifications have been unmanageable the last few weeks and this slipped through the cracks.

Since #229 preceded this and had files in places that were more similar to the rest of the project structure, I've manually merged your work into that PR, added you as a co-author there and merged it instead of this one. Thanks again!

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@2.2.0`. Thanks again!

PR #275, opened 2020-11-10
fix: blue text works better across themes instead of blue background

TylerBarnes:
Current style: 

Screen Shot 2020-11-10 at 2 51 12 PM

Screen Shot 2020-11-10 at 2 51 22 PM

New style:

Screen Shot 2020-11-10 at 2 59 36 PM

Screen Shot 2020-11-10 at 2 50 15 PM

### DIFF
--- a/plugin/src/utils/format-log-message.js
+++ b/plugin/src/utils/format-log-message.js
@@ -16,7 +16,7 @@ const formatLogMessage = (input, { useVerboseStyle } = {}) => {
   }
 
   return verbose || useVerboseStyle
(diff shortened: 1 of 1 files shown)


TylerBarnes:
I left spaces around `gatsby-source-wordpress` intentionally so that it doesn't line up with the log messages on other lines. This is more of a tag on a log message instead of the actual message output, so it felt wrong for it to perfectly line up with log message output and it was a bit harder to read. This way your eyes don't treat them as being at the same level and it's easier to mentally group together other messages vs messages from this plugin.

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@2.1.2`

Issue #151, opened 2020-09-07
imageMaxWidth not passing to image wrapper


robmarshall:
I am setting the following options in my config:

` html: {
        imageMaxWidth: 700,
},`

However, the wrapper div that surrounds the image is having its width set as the WordPress image size, not the config size. Meaning the image breaks out of the div. I am having to pass an !important css style to stop this.

See this article for example: https://thoughtsandstuff.com/deploy-gatsby-aws-with-ci-without-amplify
Repo: https://github.com/robmarshall/tns-gatsby/blob/master/gatsby-config.js

TylerBarnes:
A fix for this is published in `gatsby-source-wordpress-experimental@2.1.4`. Thanks for opening this issue @robmarshall  and apologies for the lengthy delay in getting it fixed! So many competing priorities with this plugin these days 😱

robmarshall:
@TylerBarnes Nice one! Thanks.

PR #282, opened 2020-11-12
fix: imageMaxWidth overrides maxWidth

acao:
fixes issue #151

note that this will only work with images from the same source as the configured wordpress host


TylerBarnes:
Looks great :D

TylerBarnes (approved):

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@2.1.4` thanks @acao!

acao:
thanks for pairing with me!

PR #284, opened 2020-11-13
feat: test /graphql in case provided url fails

acao:
if we get an HTML response when checking whether the schema URL is valid, we make a more minimal, successive request that tests appending `/graphql` to the URL. it either throws an error that instructs the user to update their config, or it fails quietly and goes on to handle the HTML response as before.


TylerBarnes:
Looking great 😄 thanks for this!

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@2.3.0`

PR #287, opened 2020-11-16
Bump to support WP release

seagyn:
Just a really small commit allowing the release of V1.


TylerBarnes:
Oops, didn't see this! :p I added support about 10 mins ago too 😅 thank you for this though!

PR #302, opened 2020-11-19
Fix: limit option breaks schema for non null fields

### LINKED ISSUE #192: Feature request: Non-null fields behaviour on local build
Consider the following situation:
600 posts on live site
ACF post object fields pointing to related posts

20 posts on development site
ACF post object fields pointing to non existing posts

gatsby build produces `ERROR #85925 Cannot return null for non-nullable field WpPost.uri` 

It would be convenient if the fields would become nullable in this situation.

TylerBarnes:
Closes #192 

cc @wardpeet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 3.1.1
+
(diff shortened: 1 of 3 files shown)


TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@3.1.1`

PR #304, opened 2020-11-20
Fix: add missing packages

TylerBarnes:
Gatsby core recently removed `cache-manager-fs-hash` which this plugin was importing. Unfortunately this plugin didn't have it declared as a dependency.


garryburch:
Thank you - this got us today on a merge - I reverted to lower version 3.0.1 - Thank you again for all your hard work!

PR #314, opened 2020-11-24
feature: seperate plugin options for concurrency limits, better errors

acao:
this trades out the env variable for `requestConcurrency` for two different plugin options

## TODO:
- [x] plugin options validation
- [x] docs
- [ ] override using/fallback to env variable?
- [x] seperate `requestConcurrency` for `MediaItem`
- [x] seperate `requestConcurrency` for `schema`
- [x] better errors for 5xx requests that are usually caused by request overloads

### DIFF
--- a/plugin/src/models/gatsby-api.ts
+++ b/plugin/src/models/gatsby-api.ts
@@ -40,6 +40,7 @@ export interface IPluginOptions {
     typePrefix: string
     timeout: number // 30 seconds
     perPage: number
+    requestConcurrency?: number
   }
   excludeFieldNames: []
   html: {
@@ -61,6 +62,7 @@ export interface IPluginOptions {
       localFile?: {
         excludeByMimeTypes?: string[]
         maxFileSizeBytes?: number
+        requestConcurrency?: number
       }
     }
   }
@@ -103,6 +105,7 @@ const defaultPluginOptions: IPluginOptions = {
     typePrefix: `Wp`,
     timeout: 30 * 1000, // 30 seconds
     perPage: 100,
+    requestConcurrency: 50,
   },
   excludeFieldNames: [],
   html: {
@@ -158,6 +161,7 @@ const defaultPluginOptions: IPluginOptions = {
       localFile: {
         excludeByMimeTypes: [],
         maxFileSizeBytes: 15728640, // 15Mb
+        requestConcurrency: 100,
       },
       beforeChangeNode: async ({
         remoteNode,
--- a/plugin/src/models/gatsby-api.ts
+++ b/plugin/src/models/gatsby-api.ts
@@ -105,7 +105,7 @@ const defaultPluginOptions: IPluginOptions = {
     typePrefix: `Wp`,
     timeout: 30 * 1000, // 30 seconds
     perPage: 100,
-    requestConcurrency: 50,
+    requestConcurrency: 10,
   },
   excludeFieldNames: [],
   html: {
--- a/plugin/src/models/gatsby-api.ts
+++ b/plugin/src/models/gatsby-api.ts
@@ -105,7 +105,7 @@ const defaultPluginOptions: IPluginOptions = {
     typePrefix: `Wp`,
     timeout: 30 * 1000, // 30 seconds
     perPage: 100,
-    requestConcurrency: 10,
+    requestConcurrency: 15,
   },
   excludeFieldNames: [],
   html: {
--- a/plugin/src/steps/source-nodes/create-nodes/create-remote-file-node/index.js
+++ b/plugin/src/steps/source-nodes/create-nodes/create-remote-file-node/index.js
@@ -72,36 +72,43 @@ const CONNECTION_TIMEOUT = 30000
  * When pushing a task with a similar id, prefer the original task
  * as it's already in the processing cache
  */
-const queue = new Queue(pushToQueue, {
-  id: `url`,
-  merge: (old, _, cb) => cb(old),
-  concurrent: process.env.GATSBY_CONCURRENT_DOWNLOAD || 200,
-})
+
+let queue = null
+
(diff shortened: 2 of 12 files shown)


TylerBarnes (changes requested):
This generally looks great but there is one misunderstanding in the changes. Right now GraphQL request concurrency is limited to 1 per type. The maxRPS util partially controls concurrency in that it limits how many requests can happen per second, but the concurrency is actually set up in `src/steps/source-nodes/fetch-nodes/fetch-nodes.js` on line 123 and 127. It really should be a queue instead of just chunking the requests, but this is what really controls the GraphQL request concurrency because each node type is fetched 1 page a time but in that file ☝️ we chunk out groups of each node type. So if the request concurrency was 2, we would have 2 paginated recursive fetches for Pages and Posts, then when those finish paginating and doing all their string of requests we would have 2 more paginated recursive fetches for Users and Categories and so on. Part of the reason it's like this is so that the activity timers show accurate times. If we started all at once and then used maxRPS to control concurrency all timers would start together but the requests would wait, so it would look like the sourcing requests would be slowing down as it goes through each type, but the reality would be that all timers started together but the requests didn't start until later due to maxRPS. If that makes sense 😅 haha!

TylerBarnes:
I think we need to rebase from master to get the tests to run 👍

TylerBarnes:
For the question of ` override using/fallback to env variable?`, I think let's make this a breaking change `v4.0.0` and include a note about `GATSBY_CONCURRENT_REQUEST` not being used anymore in the docs as well as in the error messages.

plugin/src/utils/report.js
```diff
@@ -5,6 +5,7 @@ export const CODES = {
   RequestDenied: `111004`,
   Authentication: `111005`,
   Timeout: `111006`,
+  WordPress500ishError: `111007`,
```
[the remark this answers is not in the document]
acao: oops!

acao:
@TylerBarnes do we want to create a troubleshooting link header for this, rather than adding more content to the errors?

TylerBarnes:
I think it's nicer for the user to not have to leave the terminal to fix it but we could also include a link to docs. Ideally we wouldn't link to docs though, especially since we'll be moving everything around soon

acao:
@TylerBarnes ok, i think it's all ready now!

plugin/src/models/gatsby-api.ts
```diff
@@ -103,6 +105,7 @@ const defaultPluginOptions: IPluginOptions = {
     typePrefix: `Wp`,
     timeout: 30 * 1000, // 30 seconds
     perPage: 100,
+    requestConcurrency: 50,
```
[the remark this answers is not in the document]
acao: done! agreed, this makes a lot of sense.

plugin/src/steps/source-nodes/create-nodes/create-remote-file-node/index.js
```diff
@@ -371,7 +378,9 @@ module.exports = ({
   ext = null,
   name = null,
   reporter,
+  pluginOptions,
 }) => {
+  const limit = pluginOptions?.type?.MediaItem?.localFile?.requestConcurrency
```
acao, line 383: i guess we technically don't need optionals here, since we're setting defaults right?

TylerBarnes: We might since a user could potentially set type.MediaItem null 🤔 unless Joi is preventing that from happening

acao:
wait... how does changing a markdown file cause the schema test suite to fail? but other test suites to pass?

plugin/src/steps/source-nodes/create-nodes/create-remote-file-node/index.js
acao: ohh right, good point

acao: better safe than sorry!

acao:
image


ok, weird to get a snapshot failure for what appears to be an upstream change??

was there a version released on top of another existing version? afaik, this should *only* happen if we change something with the schema, or a wp-graphql* plugin is bumped?

TylerBarnes:
Looks like the WPGraphQL schema changed somehow. Maybe it updated to a newer version? 🤔

TylerBarnes:
Ahhh, it's because of a recent WPGatsby change. I added a field `WPGatsby.isPreviewFrontendOnline`

acao:
just confused as to why this would happen without changing plugin versions in the PR? 👀

TylerBarnes:
If you merged or rebased master I likely updated the version elsewhere and that made it's way in here.

acao:
@TylerBarnes then the snapshots should have updated as well when I rebased??

acao:
i'll try rebasing again

TylerBarnes:
Yeah that's a good point! I'm not sure how to explain this haha. The weird thing is the docker-compose.yml is still on v0.6.0 but the change here wasn't introduced until v0.6.7. 

Maybe this is a bug with our docker setup script?

https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/docker-compose.yml#L25

acao:
@TylerBarnes it appears to be that a wordpress plugin version was bumped without updating the snapshots. Thats why every merge to master fails CI on the test schema job

TylerBarnes:
I'm confused though because on master the version is still v0.6.0 which shouldn't have this change in it

TylerBarnes:
Looks like we should use the install-plugin script instead of using git clone https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/docker/wordpress/Dockerfile#L35

TylerBarnes (approved):

PR #321, opened 2020-11-26
Fix: update schema breaks preview

### LINKED ISSUE #318: Wordpress updates cause the GraphQL schema to loose inferred types of remote nodes
V3.1.3.
To emulate:
Running Gatsby develop (as used by preview) when build completes, any remote nodes (our bookingGraph node in this case) are present on the graph and can be seen in the playground alongside the WP types.
However go to Wordpress Admin Press and update a page then a build is triggered and this results in the remote node (our bookingGraph Node) disappearing and the build fails with Cannot query field "bookingGraph" on type "Query" as this is required in our pageQuery.
Once this happens it will never recover.

**I can prevent this by removing the following from 'steps/source-nodes/update-nodes/wp-actions/update.js':**
```
  // await (0, _customizeSchema.customizeSchema)({
  //   parentSpan: null,
  //   deferNodeMutation: true,
  //   refresh: false,
  //   ...args
  // });
```

page query:

```js
export const pageQuery = graphql`
  query($id: String!, $propId: Int! ) { 
    wpProperty(id: { eq: $id }) {
      id
      content
     ....other stuff
    }
    bookingGraph {
      getProperty(id: $propId) {
        ...propertyFragments
      }
    }
...... other stuff
}
```

gatsby.config.js excerpt:
```js
    {
      resolve: "gatsby-source-graphql",
      options: {
        typeName: "BookingGraph",
        fieldName: "bookingGraph",
        url: `${process.env.GATSBY_GRAPH_ADDR}/graphql`,
      },
    },
```
----

TylerBarnes:
Fixes #318 

In addition, I've made adjacent changes related to this issue to make sure our DX doesn't degrade by fixing this.

## 3.3.0

- Version 3.0.0 made an attempt to remove expensive schema diffing at the beginning of each preview or delta update. The strategy was that when a GraphQL query to WPGraphQL failed, it would try to re-run createSchemaCustomization internally to recover and rebuild our internal sourcing queries before trying to make the query again. Unfortunately this caused issues because more recent versions of Gatsby do not allow createSchemaCustomization to be called in this way. This version reverts to the earlier behaviour where the remote schema md5 is diffed on every build or preview.
- Because of the above issue, automatic schema updates during `gatsby develop` also failed to keep working. We now use the Gatsby refresh API to re-trigger node sourcing and schema customization in our develop watcher as this is a public API and will be more stable than Gatsby internals.
- `ensurePluginRequirementsAreMet` is called once on a cold cache but never again on a warm cache. We now call this again when the schema changes on a warm cache to make sure that if the schema changes because the remote plugin versions changed, we recheck to make sure they're still in-range.
- The default value of `options.develop.nodeUpdateInterval` has been increased from `300` to `5000`. Due to the time it takes everything to work across Gatsby and WP, as well as the fact that content is rarely updated at the exact start of the interval, this change is barely perceptible but will ease up some strain on WP when many developers are working on the same site or when `gatsby develop` is running for hours at a time.

### DIFF
--- a/plugin/src/steps/source-nodes/update-nodes/content-update-interval.js
+++ b/plugin/src/steps/source-nodes/update-nodes/content-update-interval.js
@@ -64,9 +64,9 @@ const refetcher = async (
   { reconnectionActivity = null, retryCount = 1 } = {}
 ) => {
   try {
-    const { pauseRefreshPolling } = store.getState().develop
+    const { refreshPollingIsPaused } = store.getState().develop
 
-    if (!pauseRefreshPolling) {
+    if (!refreshPollingIsPaused) {
       await checkForNodeUpdates(helpers)
     }
 
--- a/plugin/src/steps/source-nodes/update-nodes/content-update-interval.js
+++ b/plugin/src/steps/source-nodes/update-nodes/content-update-interval.js
@@ -1,7 +1,62 @@
-import fetchAndApplyNodeUpdates from "./fetch-node-updates"
 import { formatLogMessage } from "~/utils/format-log-message"
 import store from "~/store"
 import { getGatsbyApi } from "~/utils/get-gatsby-api"
+import { contentPollingQuery } from "../../../utils/graphql-queries"
+import fetchGraphql from "../../../utils/fetch-graphql"
+import { LAST_COMPLETED_SOURCE_TIME } from "../../../constants"
+
+const previouslyFoundActionIds = new Set()
+
+/**
+ * This function checks wether there is atleast 1 WPGatsby action ready to be processed by Gatsby
+ * If there is, it calls the refresh webhook so that schema customization and source nodes run again.
+ */
+const checkForNodeUpdates = async ({ cache, emitter }) => {
+  // get the last sourced time
+  const since = await cache.get(LAST_COMPLETED_SOURCE_TIME)
+
+  // make a graphql request for any actions that have happened since
+  const {
+    data: {
+      actionMonitorActions: { nodes },
+    },
+  } = await fetchGraphql({
+    query: contentPollingQuery,
+    variables: {
+      since,
+    },
+    // throw fetch errors and graphql errors so we can auto recover in refetcher()
+    throwGqlErrors: true,
+    throwFetchErrors: true,
+  })
+
+  // a very simple cache to make sure we don't process the same action twice within
+  // the same `gatsby develop` process.
+  const newActions = nodes.filter(({ id }) => {
+    if (previouslyFoundActionIds.has(id)) {
+      return false
+    } else {
+      previouslyFoundActionIds.add(id)
+      return true
+    }
+  })
+
+  if (newActions.length) {
+    // if there's atleast 1 new action, pause polling,
+    // refresh Gatsby schema+nodes and continue on
+    store.dispatch.develop.pauseRefreshPolling()
+
+    emitter.emit(`WEBHOOK_RECEIVED`, {
+      webhookBody: {
+        since,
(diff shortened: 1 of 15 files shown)


TylerBarnes:
@acao I'm not sure why but the tests are failing at the docker part 🤔 weird since it's been working perfectly for weeks

acao:
@TylerBarnes yep looks like a bug with fetching the npm cache. try rerunning

TylerBarnes:
I re-ran it a few times and it fails every time 🤔 maybe for this one I'll just run the tests locally and we can sort out the tests on Monday. Thanks @acao !

acao:
oh the issue is with a script. did someone modify a Dockerfile? or is there a release not available via tar gz? imo we should address the error message first

TylerBarnes:
As far as I'm aware no one modified the docker setup. By releases do you mean for our WP dependencies?

acao:
yes, the log message shows it’s failing on curl’ing a wp plugin, i can’t tell which until i get home

TylerBarnes:
Looks like it's failing when unzipping an archive.zip file 🤔 

```
Step 23/28 : RUN unzip -a -d ${WP_CONTENT_DIR} /tmp/archive.zip
 ---> Running in 5fe7fc5c7ec3
Archive:  /tmp/archive.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /tmp/archive.zip or
        /tmp/archive.zip.zip, and cannot find /tmp/archive.zip.ZIP, period.
```

TylerBarnes:
No worries, there's no rush on it since I'm able to run tests locally anyway. Enjoy your holiday!

TylerBarnes:
Ahhh, nevermind I can't run locally either. Looks like our uploads zip went down. https://storage.googleapis.com/gatsby-source-wordpress-v4-wordpress-test-files/gatsby-source-wordpress-test-uploads.zip

I can get that sorted 👍 thanks @acao !

plugin/src/steps/source-nodes/update-nodes/content-update-interval.js
```diff
@@ -69,22 +124,17 @@ const refetcher = async (
...
-  const { verbose } = store.getState().gatsbyApi.pluginOptions
+  const { verbose, develop } = store.getState().gatsbyApi.pluginOptions
 
-  const msRefetchInterval =
-    pluginOptions &&
-    pluginOptions.develop &&
-    pluginOptions.develop.nodeUpdateInterval
-      ? pluginOptions.develop.nodeUpdateInterval
-      : 300
+  const msRefetchInterval = develop?.nodeUpdateInterval ?? 300
```
smthomas, line 137: Should this value default to 5000 like in the other locations? 

Nit: Is there a way to re-use that value so the default doesn't need to be updated in multiple places?

TylerBarnes: Ahh good point :) actually it doesn't need to have a fallback here as the redux store has a default already. I'll remove this entirely

TylerBarnes: And we could use a constant to tie it all together 🤔 we should probably do that for all of the default values actually. Maybe we should do that in a separate PR

smthomas (approved):

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@3.3.0`

PR #325, opened 2020-11-27
fix: resolve aliased paths properly with typescript

acao:
this helps with:

- moving the project into gatsby monorepo
- setting up baseline typescript support (need at least some basic config)
- getting imports working in vscode and other tools, and with inference engine

notes:
- this actually compiles with tsc! before, tsc would fail when checking these paths
- gatsby's babel config doesn't type check, so watch out! always try `tsc` from the `plugin` directly to confirm valid typescript types, unlike the usual build

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
     "babel-eslint": "^10.1.0",
     "babel-jest": "^25.1.0",
     "babel-plugin-import-globals": "^2.0.0",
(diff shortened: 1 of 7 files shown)


acao:
oops! forgot i need to set up jest to work with typescript!

acao:
I think I will close this, and move this effort into the typescript conversion effort.

Issue #310, opened 2020-11-23
pathPrefix missing for local links inside content


henrikwirth:
Inside content links are missing the pathPrefix in version 1.7.9.

Since I don't see anything related to it in the changelog, I suspect this is still a problem in newer versions.

@TylerBarnes Is there any way we could spin up a 1.7.10 or 1.8. only with that fix?  I'm stuck on these version since I can't upgrade WPGatsby or WPGraphQL any further for now.

henrikwirth:
It seems this could be a problem of pathPrefix not being passed down to the `gatsby-plugin-catch-links`.

Is there a way to overwrite the plugin and use it in our main project instead?

TylerBarnes:
@henrikwirth you should be able to add `gatsby-plugin-catch-links` to your `gatsby-config.js` and it'll use that instead.

TylerBarnes:
I don't see an option for that in that plugin though, so this plugin may need to check for a path prefix and add it to the inline relative links

TylerBarnes:
Yep, looks like this plugin needs to handle this https://github.com/gatsbyjs/Gatsby/blob/master/packages/gatsby-plugin-catch-links/src/catch-links.js#L167

henrikwirth:
So after playing around a little, I figured, that the problem mainly arises from the fact, that the links inside content are written as follows:

```
<a href="/some-internal-link/">Some Link</a>
```

The problem is that usually, if you would use GatsbyLink on a link like this, it would prefix the link automatically. This doesn't happen for the links inside Wysiwyg though, since the node process function only picks up links, that have the actual WordPress backend domain in front. Like:

```
<a href="https://wp-backend.com/some-internal-link/">Some Link</a>
```

I think it would be great to have that as an option. like `add path prefix to relative urls inside content`. And I believe this should even be the default behavior.

TylerBarnes:
I'm cleaning up this repo a bit and moving this issue to our internal backlog. If anyone finds this issue and needs this sooner please add a feature request to https://portal.gatsbyjs.com/gatsby-cloud

PR #328, opened 2020-11-29
fix: relative urls inside html content to include prefix

henrikwirth:
Issue: #310 

So when using pathPrefix, relativeUrls inside html don't get prefixed. This doesn't make much sense, since I might want to create content in WordPress with relative URLs that don't use the prefix yet, so later I can change the prefix to whatever I want.

Where this comes in handy and was really necessary, is a multisite setup with multiple translations. 

What this fix does: When pathPrefix is used and the `pluginOption.prefixRelativeUrls` is active, the nodeStrings will be searched for elements, that start with `href=` and include a url that starts with `/`.

Excluded by the regex are urls that start with: `pathPrefix|/wp-content|/wp-admin|/wp-includes|//`

So we don't want prefix already prefixed URLs or URLs that start with a double `//`. 

The other branch that was branched from v1.7.9 can be found here since I need a v1.7.10: https://github.com/henrikwirth/gatsby-source-wordpress-experimental/tree/fix-relative-urls-in-html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,10 @@ dist
 
 .DS_Store
 yarn-error.log
+
+# IDEs
 .vscode
+.idea
 
 node_modules
 public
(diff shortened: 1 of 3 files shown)


henrikwirth:
Maybe I could leave `/wp-content|/wp-admin|/wp-includes` outside of the regex, since we are not looking for `wpUrl` in that regex.

PR #332, opened 2020-12-01
Feat: typescript conversion

TylerBarnes:
@acao I had a quick crack at this and converted all the util/*.js files :)

accomplishes:

- typescript ready file aliases that make vscode happy
- converting utils, store and some steps files as an example
- types for store state (no types for store actions yet, hmm...)
- sprinkle `GatsbyNodeAPI`  and `Reporter` types throughout
- address a slight type def bug with `debug.timeBuildSteps`
- add more IDs to reporter messages


acao:
@TylerBarnes is it ok of i commit to this branch? i'll assume yes

TylerBarnes:
@acao yes please do 😄

TylerBarnes:
Oh! I just remembered rematch has really good typescript support for our data stores/models. I set it up on another project and ts is able to infer types from calls to getState() or dispatch() which is really nice. I'll have to add that here too

acao:
@TylerBarnes that sounds nice, but doesn't sound like the typescript way to do things. type definitions should always ship with their modules if possible, according to the docs

TylerBarnes:
Definitely 👍 The types do ship with the module, I just mean rematch carries them through so when we're using it it's set up to properly use our types.

acao:
@TylerBarnes ohh i see what you mean, with automatic inference. yes i'm working on setting that up in the typescript definitions

acao:
right now, my goal is just to get a proper tsc build without typechecking errors. it seems the gatsby babel config is configured to skip typescript compilation errors which is really dangerous, because you can end up shipping broken module versions very easily.

TylerBarnes:
If we use tsc directly without babel wont our compiled code be different? Since we're using babel to enable different features. Or do we need to run them both?

acao:
@TylerBarnes we should not use TSC to actually build the project, but the fact of the matter is that the compiler is failing to properly compile typescript, which could lead to missing files. tsc is just the only way for me to manually check for compiler errors, since babel is configured to skip over them for now

TylerBarnes:
Ahh, got it :)

acao:
image

I'm confused why these are marked as "not typed by gatsby". they just aren't available in a cohesive interface of course, but most of what we use is available!

acao:
store types were as easy as changing the store.js file to store.ts!

acao:
this part got even easier on some further digging

image

plugin/src/utils/cache.ts
```diff
@@ -337,7 +343,7 @@ export const restoreHardCachedNodes = async ({
       }
 
       // restore each node
-      await helpers.actions.createNode(remoteNode)
+      helpers.actions.createNode(remoteNode)
```
acao, line 346: I have a theory that, depending on how we are transpiling async/await, that this would introduce a performance improvement. typescript revealed this was an unnecessary use of await

TylerBarnes: nice!

plugin/src/utils/report.ts
```diff
@@ -15,49 +17,58 @@ export const CODES = {
...
   },
   [CODES.SourcePluginCodeError]: {
-    text: (context) => context.sourceMessage,
+    text: getErrorText,
     level: `ERROR`,
     category: `SYSTEM`,
   },
   [CODES.WordPress500ishError]: {
-    text: (context) => context.sourceMessage,
+    text: (context: IErrorContext): string => context.sourceMessage,
```
acao, line 71: oops, i missed this when pulling in the upstream commits

plugin/src/utils/cache.ts
TylerBarnes: I think the core type on this is probably wrong 🤔 or TS is wrong somehow. If you check the return type in the actions docs it returns a promise which resolves when all cascading createNode calls are finished https://www.gatsbyjs.com/docs/actions/#createNode

package.json
```diff
@@ -70,7 +70,7 @@
       "eslint",
       "prettier --write"
     ],
-    "*.{md,json,yml}": [
+    "*.{json,yml}": [
```
TylerBarnes, line 73: Don't we want to format our docs still?

plugin/src/hooks/node-filters.ts
```diff
@@ -1,5 +1,21 @@
...
+  name: string
+  context: JSON
+  data: JSON
+}
+
+type NodeFilterFn = (INodeFilterInput) => JSON
+
+interface INodeFilter {
+  name: string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
TylerBarnes, line 14: looks like there's no any here now :)

plugin/src/models/gatsby-api.ts
```diff
@@ -83,6 +87,7 @@ export interface IPluginOptions {
     requestConcurrency?: number
   }
   excludeFieldNames?: []
+  excludeFields?: string[] // deprecated?
```
TylerBarnes, line 90: Yeah, this is deprecated and renamed to the above `excludeFieldNames` which is an array of strings.

plugin/src/models/logger.ts
```diff
@@ -1,14 +1,32 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
TylerBarnes, line 2: Do we still need these eslint-disable comments?

plugin/src/steps/check-plugin-requirements.ts
```diff
@@ -53,6 +58,7 @@ const areRemotePluginVersionsSatisfied = async ({
 
   try {
     const { data } = await fetchGraphql({
+      url: wpGraphQLEndpoint,
```
TylerBarnes, line 61: Do we need to add this? I thought `fetchGraphql` can access the url internally. I think we only need to pass the url explicitly for our tests or anywhere that we're making a request to an API besides WPGraphQL.

plugin/src/steps/check-plugin-requirements.ts
```diff
@@ -206,8 +218,9 @@ ${firstError.debugMessage || firstError.message}
     })
     .catch((e) => {})
 
-const isWpGatsby = async () =>
+const isWpGatsby = async ({ url }: { url: string }) =>
   fetchGraphql({
+    url,
```
TylerBarnes, line 223: same as above :)

plugin/src/steps/check-plugin-requirements.ts
```diff
@@ -222,9 +235,16 @@ const isWpGatsby = async () =>
     isFirstRequest: true,
   })
 
-const prettyPermalinksAreEnabled = async ({ helpers }) => {
+const prettyPermalinksAreEnabled = async ({
+  helpers,
+  url,
```
TylerBarnes, line 240: And this one :D

plugin/src/steps/ingest-remote-schema/diff-schemas.js
```diff
@@ -35,7 +34,7 @@ const checkIfSchemaHasChanged = async ({ traceId }) => {
   }
 
   const { data } = await fetchGraphql({
-    query: gql`
+    query: /* GraphQL */ `
```
TylerBarnes, line 37: Good call. We should delete `utils/gql` if you didn't already. It's not really needed anywhere and doesn't do more than a comment

plugin/src/utils/cache.ts
```diff
@@ -125,7 +142,11 @@ export const setHardCachedData = async ({ key, value }) => {
   await hardCache.set(key, value)
 }
 
-export const getHardCachedData = async ({ key }) => {
+export const getHardCachedData = async <T = Node>({
```
TylerBarnes, line 145: This isn't necessarily a node but can be any cacheable data (so a string, number, object, node, array, etc)

plugin/src/utils/cache.ts
```diff
@@ -134,11 +155,11 @@ export const getHardCachedData = async ({ key }) => {
...
   const data = await hardCache.get(key)
 
-  return data
+  return data as T
 }
 
-export const getHardCachedNodes = async () => {
-  const allWpNodes = await getHardCachedData({ key: `allWpNodes` })
+export const getHardCachedNodes = async (): Promise<null | Node[]> => {
+  const allWpNodes = await getHardCachedData<Node[]>({ key: `allWpNodes` })
```
TylerBarnes, line 162: Same as above :)

plugin/src/utils/cache.ts
```diff
@@ -15,30 +15,41 @@ import {
 } from "~/steps/create-schema-customization/helpers"
 
 import { createMediaItemNode } from "~/steps/source-nodes/fetch-nodes/fetch-referenced-media-items"
+import { GatsbyNode, Node } from "gatsby"
```
TylerBarnes, line 18: Looks like `GatsbyNode` isn't being used anymore

plugin/src/utils/cache.ts
```diff
@@ -305,24 +343,27 @@ export const restoreHardCachedNodes = async ({ hardCachedNodes }) => {
       }
 
       // restore each node
-      await helpers.actions.createNode(remoteNode)
+      helpers.actions.createNode(remoteNode)
```
TylerBarnes, line 346: I mentioned it in your comment but just adding a reminder here too that we should await this

plugin/src/utils/fetch-graphql.ts
```diff
@@ -488,6 +540,7 @@ if ( defined( 'GRAPHQL_REQUEST' ) && true === GRAPHQL_REQUEST ) {
   }
 
   reporter.panic({
+    id: CODES.BadResponse,
```
TylerBarnes, line 543: Good catch on these! 😄

plugin/src/utils/gatsby-types.ts
```diff
@@ -1,2 +1,7 @@
+// these aren't typed in Gatsby..
+
+import type { NodePluginArgs, Reporter } from "gatsby"
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type GatsbyHelpers = any
+export type GatsbyNodeApiHelpers = NodePluginArgs
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
TylerBarnes, line 6: Can we remove the comments now? :)

plugin/src/utils/gql.js
```diff
@@ -1,3 +0,0 @@
-const gql = ([string]) => string
-
-export default gql
```
TylerBarnes, line 3: Ah nice you did delete it, glad to be rid of this :D

plugin/src/utils/report.ts
TylerBarnes, line 71: Just adding a reminder in my review :) we should probably use the `getErrorText` function on this one as well. nice clean up here!

plugin/src/utils/run-steps.ts
```diff
@@ -1,15 +1,28 @@
...
+): Promise<void> => {
   for (const step of steps) {
     try {
       const { timeBuildSteps } = pluginOptions?.debug ?? {}
       const timeStep =
         typeof timeBuildSteps === `boolean`
           ? timeBuildSteps
-          : timeBuildSteps?.includes(step.name) ||
+          : // TODO: type for array option? or deprecated setting?
+            timeBuildSteps?.includes(step.name) ||
```
TylerBarnes, line 25: This was meant to be an internal API for debugging the source plugin really. It allows timing all build steps by setting it true or timing a specific set of steps by including their names as an array.

plugin/src/utils/run-steps.ts
```diff
@@ -58,10 +71,14 @@ const runSteps = async (steps, helpers, pluginOptions, apiName) => {
...
-  runSteps(steps, helpers, pluginOptions, apiName)
+const runApiSteps = (steps: Step[], apiName: string) => async (
+  helpers: GatsbyNodeApiHelpers,
+  pluginOptions: IPluginOptions
+): Promise<void> => runSteps(steps, helpers, pluginOptions, apiName)
 
-const runApisInSteps = (nodeApis) =>
+const runApisInSteps = (nodeApis: {
+  [apiName: string]: Step | Step[]
+}): { [apiName: string]: Promise<void> } =>
```
TylerBarnes, line 81: This actually doesn't return void :) It returns a Gatsby config object with node api names as keys and a function as a value which runs all the steps that are passed into it.

plugin/tsconfig.json
```diff
@@ -1,8 +1,10 @@
 {
+  "extends": "../tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
+    "baseUrl": "./src",
     "paths": {
-      "~/*": ["src/*"]
+      "~/*": ["*"]
```
TylerBarnes, line 6: Thank goodness haha! Does this mean we can command click on imports again?? :D

TylerBarnes (commented):
Overall looking really great, awesome work @acao ! So stoked on this :D it'll be so much easier to work on these parts of the codebase going forwards. I left just a small handful of questions and changes but we should be able to merge this really soon :D

Since I'm the original author it wont let me request changes 😆 so this is just a set of comments I guess. But pretend it's a change request?

plugin/src/steps/ingest-remote-schema/diff-schemas.js
TylerBarnes: Ah, nice, you did delete it!

plugin/src/utils/run-steps.ts
TylerBarnes: Oh 🤦 lol that's what you typed it as! Ignore this :)

plugin/src/utils/gql.js
acao: yes i wasnt sure what it was? sometimes folks do this because the LSP server only offers language support for `gql` tag right now. hoping to find the free time to fix that soon!

plugin/src/utils/report.ts
acao: ah yes, i noticed that, which is what i referred to here:

https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/pull/332#discussion_r542813510

happened because of an upstream rebase with the primary branch

plugin/src/utils/run-steps.ts
acao: ahh ok! well currently in vscode it says that `pluginOptions.debug.timeBuildSteps` does not exist. `IPluginOptions` doesn't have this field, should I add it?

plugin/tsconfig.json
acao: yes!! seems that way so far. and it also works with automatic local module imports too 💯

plugin/src/models/gatsby-api.ts
acao: i think there's a way to annotate this in jsdoc with `@deprecated` which will surface some new vscode features (autocomplete notes, rendering a strike through for deprecated methods/etc!)

plugin/src/models/logger.ts
acao: good catch, no we do not because i addressed this in the config itself

plugin/src/steps/check-plugin-requirements.ts
acao: ah! ok so we need to make `url` optional then? this is one other runtime change i forgot about!

plugin/src/utils/run-steps.ts
acao: @TylerBarnes ah i see the issue here! so `debug.timeBuildSteps` is listed as a `boolean` in `IPluginOptions`. it should be an array of strings instead?

acao: https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/pull/332/commits/cdd8d75c56e9cf71c6fc472e6860c3bcea219c1a

TylerBarnes: It can be either an array of strings or a boolean

acao: can it just be an empty array? i guess that would be a breaking change

acao: ohhhh i see

plugin/src/utils/cache.ts
acao: ok! i can just tsignore it for now. it doesn't really impact any inference outside itself

plugin/src/models/gatsby-api.ts
acao: image

TylerBarnes: Folks have been warned about this for the past 6 or 7 months https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/plugin/src/steps/process-and-validate-plugin-options.js#L19

So I think we can remove it

package.json
acao: yeahhh :P

acao:
last build failed because raw.githubusercontent.com was down? 👀

plugin/src/utils/fetch-graphql.ts
acao: ah yes, the gatsby definitions enforced this actually! the typescript error is hard to read, but i found that logging `panic({ context: {})` was invalid, wnd adding the ID made it valid. i think it has to do with the discrepancy in how typescript treats type unions vs function overloads. i'm making a little list of core typescript improvements to make for this!

plugin/src/utils/cache.ts
acao: yeah, i make it a gatsby node by default because most of the cases were for returning `Node[]` or `node`, but the type arg is overridable, so you can provide whatever response type when casting

plugin/src/utils/cache.ts
acao: yup! thats why i provided the type arg, since in this case it's returning nodes. you can provide `getHardCachedData<string>` or whatever you expect

plugin/src/utils/cache.ts
```diff
@@ -4,7 +4,7 @@ import fsStore from "cache-manager-fs-hash"
 import path from "path"
 import rimraf from "rimraf"
 
-import store from "~/store"
+import { default as pluginStore } from "~/store"
```
acao, line 7: i did this to prevent collisions below with `store` module and `store` as an argument, but that seems to be causing errors now i think?

TylerBarnes: huh 🤔 maybe change it back? It's definitely working as-is. I think there wont be a collision since the local scope will override the module scope.

acao:
to note: ctrl+click on files *only* works with other typescript files currently, but it works!  and most importantly, it will still validate paths for javascript, and ctrl-click still works on module imports

for e.g:

image

it recognizes the typo in the filename which is actually plural, ctrl+click on the filename doesn't work, but ctrl+click/peek def on the actual module imports *does* work!


i think `checkJS` tsconfig setting could possibly enhance this and make it work for all files, but its very noisy and often inaccurate for JS (because everything is inferred from usage, it's just messy lol), which is why i didn't enable it in this PR

plugin/src/utils/cache.ts
acao: hmm... that did not fix the issue, ill keep looking!

acao: i know re-using variable names in this case doesn't break things because of scope, but it *does* drive me crazy, on principle 😆

TylerBarnes: 😆 fair enough, haha. I don't like it much either. It can make things pretty confusing! This code was copy/pasted from core to add a hard caching layer that works similar to the core caching layer. So I didn't change very much code in it ;)

plugin/src/steps/check-plugin-requirements.ts
acao: just saw these review comments now for some reason? they must have appeared after i was done for the day, oops! the above you refer to is gone but i know what you're referring to

plugin/src/utils/gatsby-types.ts
acao: yes! why are these comments only appearing in files changed view but not conversation view? so confused? github is gettin' real buggy for me, lol

acao:
image

trying to track down this bug now. definately more runtime changes than i remember :(

plugin/src/utils/cache.ts
acao: ok! keeping it as is, in terms of runtime impacting code, then.

acao:
resolved, reverted how models were exported at the root. this should be good to go now! glad to do another review pass if you prefer that

TylerBarnes:
Looks like some change here is causing problems with data hard caching in development 🤔 

Screen Shot 2020-12-16 at 9 23 55 PM

plugin/src/utils/cache.ts
```diff
@@ -15,30 +15,41 @@ import {
...
-  constructor({ name = `db`, store = fsStore } = {}) {
-    this.name = name
-    this.store = store
+  private store: Store
+  private name: string
+  private cacheDirectory: string
+  private cache: manager.MultiCache
+  constructor(options?: CacheOptions) {
+    this.name = options?.name || `db`
+    this.store = options?.store || fsStore
```
acao, line 37: @TylerBarnes this is one of the few changes to execution code left. pretty significant change. i had trouble recreating this default arg pattern in typescript for some reason but I can try to eliminate execution changes here entirely to see if that helps. it would make sense if this is it, since it deals with the hard fs cache

acao: ah, this is why it didn't work before.

before I'd tried this
image

so instead, i tried this ✅ 
image

plugin/src/utils/fetch-graphql.ts
```diff
@@ -361,13 +408,13 @@ ${slackChannelSupportMessage}`
     })
   }
 
-  const responseReturnedHtml = response?.headers[`content-type`].includes(
+  const responseReturnedHtml = !!response?.headers[`content-type`].includes(
```
acao, line 411: runtime impacting change

plugin/src/utils/fetch-graphql.ts
```diff
@@ -391,7 +438,7 @@ ${slackChannelSupportMessage}`
 
         // if adding `/graphql` works, panic with a useful message
         reporter.panic({
-          id: CODES.missingAppendedPath,
+          id: CODES.MissingAppendedPath,
```
acao, line 441: runtime impacting change

plugin/src/utils/fetch-graphql.ts
```diff
@@ -478,6 +529,7 @@ if ( defined( 'GRAPHQL_REQUEST' ) && true === GRAPHQL_REQUEST ) {
       reporter.warn(formatLogMessage(sharedEmptyStringReponseError))
     } else {
       reporter.panic({
+        id: CODES.BadResponse,
```
acao, line 532: runtime impacting change

plugin/src/utils/fetch-graphql.ts
acao, line 543: runtime impacting change

acao:
> Looks like some change here is causing problems with data hard caching in development thinking On master it's working

@TylerBarnes just made another commit to the fs cache, to revert to use the *exact* same defaults pattern as before. does that fix it?

acao:
OHHHH i remember this now. theres somewhere where a gatsby core provided interface required `owner` property!

plugin/src/utils/cache.ts
```diff
@@ -263,6 +299,7 @@ export const restoreHardCachedNodes = async ({ hardCachedNodes }) => {
       node.internal = {
         contentDigest: node.internal.contentDigest,
         type: node.internal.type,
+        owner: node.internal.owner,
```
acao, line 302: @TylerBarnes here is our culprit!

acao:
just a reminder @TylerBarnes over the holidays that this is good to review and fixes the issue you mentioned!

TylerBarnes:
For sure 👍 I'm looking at merging it right now. Just merging in some other preview changes and making sure everything is good. I'll be out til Jan 2 so see you then :D

acao:
formatting!

TylerBarnes:
Looks like some eslint stuff actually. I'm about to make a commit :)

acao:
thank you!

PR #334, opened 2020-12-01
fix: wp-gatsby should checkout tags

acao:
this will fix today's issues with snapshots in CI

### DIFF
--- a/docker/wordpress/Dockerfile
+++ b/docker/wordpress/Dockerfile
@@ -32,7 +32,7 @@ COPY install-plugin.sh /usr/local/bin/install-plugin
 RUN install-plugin wp-graphql wp-graphql/wp-graphql ${WPGRAPHQL_VERSION}
 
 # Install wp-gatsby using git, and apply a diff
-RUN git clone https://github.com/gatsbyjs/wp-gatsby ${PLUGIN_DIR}/wp-gatsby && git checkout ${WPGATSBY_VERSION}
+RUN git clone --depth 1 -b ${WPGATSBY_VERSION} https://github.com/gatsbyjs/wp-gatsby.git ${PLUGIN_DIR}/wp-gatsby 
 
 # WP GraphQL ACF from github release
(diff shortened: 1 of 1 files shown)


TylerBarnes:
Looks like there may be some other changes needed. Tests are failing with `Cloning into '/usr/src/wordpress/wp-content/plugins/wp-gatsby'...
fatal: not a git repository (or any parent up to mount point /var/www)`

docker/wordpress/Dockerfile
```diff
@@ -32,7 +32,7 @@ COPY install-plugin.sh /usr/local/bin/install-plugin
 RUN install-plugin wp-graphql wp-graphql/wp-graphql ${WPGRAPHQL_VERSION}
 
 # Install wp-gatsby using git, and apply a diff
-RUN git clone https://github.com/gatsbyjs/wp-gatsby ${PLUGIN_DIR}/wp-gatsby
+RUN git clone https://github.com/gatsbyjs/wp-gatsby ${PLUGIN_DIR}/wp-gatsby && git checkout ${WPGATSBY_VERSION}
```
TylerBarnes, line 35: I think we need to cd into the directory before checking out the tag

TylerBarnes (changes requested):

docker/wordpress/Dockerfile
TylerBarnes: Or we can clone the tag directly https://stackoverflow.com/a/21699307/12777907

TylerBarnes: --depth=1 is probably a good idea anyway :)

acao:
@TylerBarnes yep! already committed the fix :D

docker/wordpress/Dockerfile
acao: i think i already resolved this, i just pushed the commit as you were reviewing! oops haha

TylerBarnes: Perfect :D

TylerBarnes:
Woo! Thanks for sorting this out so quickly

PR #336, opened 2020-12-02
Feat: add cli that uses a list of users to run a preview swarm

TylerBarnes:
This CLI is run like `wpgatsby preview swarm --wp-url="http://previewtest.local" --users="./users.json"`. It accepts a list of users from that JSON file and then proceeds to start swarming the WP/Preview instances by repeatedly simultaneously creating new posts and previewing in many different headless chrome windows.


KyleAMathews:
EMBRACE THE CHAOS!!!

acao:
I am able to re-create this in linux locally, and now am seeing no 50x errors or PHP errors from the wordpress docker image, only 200 and 302, which makes me still wonder if i have the network properly configured

TylerBarnes:
@acao you shouldn't get any 50x errors since the preview failure on linux is due to the webhook not being hit. So I'm thinking it's probably still the network 🤔 if we're still totally stuck on this we can put it on hold for a while until we can find someone that's dealt with these cross platform networking issues before. This would be really nice to have but isn't a showstopper.

acao:
@TylerBarnes would you like me to update the PR as well?

TylerBarnes:
@acao I think we're good for now. I disabled the e2e test and added an `@todo` there. This is useable as a cli for debugging multi-user preview issues so it's valuable as-is and we can revisit getting it working in CI in the future

acao:
ahh i see ok, yes that sounds fine to me!

acao (approved):
reconciled with upstream, looks great! we will automate this in CI when we get time for it. just need to move gatsby to it's own container

TylerBarnes:
Sounds good :) thanks for the review and reconciliation @acao!

PR #344, opened 2020-12-04
Fix: missing inline html image nodes

TylerBarnes:
MediaItem nodes fetched while converting html  tags to Gatsby images were not always being properly cached. 
This would manifest as a few cryptic bugs when some other parts of the code were expecting that these nodes had been cached. This would break previews which are using html images.


plugin/src/steps/preview/index.ts
```diff
@@ -9,7 +9,6 @@ import store from "~/store"
 import { fetchAndCreateSingleNode } from "~/steps/source-nodes/update-nodes/wp-actions/update"
 import { formatLogMessage } from "~/utils/format-log-message"
 import { touchValidNodes } from "../source-nodes/update-nodes/fetch-node-updates"
-import { getDbIdFromRelayId } from "../source-nodes/update-nodes/wp-actions/update"
```
jasonbahl, line 12: I don't see any other changes in this file. Was this import just not used?

plugin/src/steps/source-nodes/create-nodes/process-node.js
```diff
@@ -147,9 +148,24 @@ const fetchNodeHtmlImageMediaItemNodes = async ({
 
         sourceUrl = ensureSrcHasHostname({ wpUrl, src: sourceUrl })
 
+        const existingNode = helpers.getNode(id)
+
+        if (!existingNode) {
+          if (!displayedFailedToRestoreMessage) {
+            helpers.reporter.warn(
+              formatLogMessage(
+                `File node failed to restore from cache. This is a bug in gatsby-source-wordpress. Please open an issue so we can help you out :)`
```
jasonbahl, line 157: totally random, but how does Gatsby handle translations for strings like this? Is there something similar to WordPress `__()` functions? Or has Gatsby not gotten that far?

jasonbahl (approved):
2 comments, but approved. LGTM

plugin/src/steps/preview/index.ts
TylerBarnes: Yeah I just noticed it was hanging around :p

plugin/src/steps/source-nodes/create-nodes/process-node.js
TylerBarnes: Yeah, Gatsby hasn't gotten that far yet

TylerBarnes:
Thanks! Published in `gatsby-source-wordpress-experimental@4.0.1`

PR #345, opened 2020-12-04
Feat: Add plugin options presets api and add a preset for preview

TylerBarnes:
This PR adds a plugin options presets api and adds a default preset for previews (to speed them up)

The default options for previews are:

```js
{
    html: {
      useGatsbyImage: false,
      createStaticFiles: false,
    },
    type: {
      __all: {
        limit: 50,
      },
      Comment: {
        limit: 0,
      },
      Menu: {
        limit: null,
      },
      MenuItem: {
        limit: null,
      },
      User: {
        limit: null,
      },
    },
  }
```

This turns off static asset processing in html fields, removes all comments, and limits all node types to 50 (with the exception of menus, menu items, and users).


jasonbahl:
So if it's only pulling say 50 categories, and the page I'm previewing has breadcrumbs or something like:

Home > Category51 > Category52 > Post 

Would that be broken because those nodes weren't fetched?

Or would these categories still be fetched when the specific preview was being viewed?

Just trying to think through cases where this will feel like a bug and not a feature.

jasonbahl:
Obviously the breadcrumbs are a hypothetical. It could be a page like a newspaper homepage that's a mix of terms and the posts within those terms and if only 50 terms were fetched but the homepage displays 

Category 51
 - Post of cat 51


Or whatever. . .just want to make sure these things will still work

TylerBarnes:
That's a good point. There's a chance it wont display them, but they can always override the settings to do something like `Tag: { limit: null }` for example. The benefit here is previews will be massively faster and more reliable by default.

jasonbahl:
Sounds good. I think it's a good default tradeoff. If we get a lot of bug reports, etc we can re-think the defaults or rethink how we surface the docs to users, etc

TylerBarnes:
I added a test and a message in the terminal explaining that it's set and how to disable it

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@5.0.0`

PR #357, opened 2020-12-13
ci: install wp-graphql* plugins on container run rather than image build

acao:
- move plugin versions to `.env.plugins` file (in preperation for further cross-repo CI automation)
- install plugins on container start rather than image build.
  - this will allow us to pre-build our images in the future which will speed up CI runs by as much as 1 minute.
- fix package resolutions for plugins (lerna publish takes care of this!)
- docs tweak


test-site/.env.plugins
```diff
@@ -0,0 +1,6 @@
+WORDPRESS_BASIC_AUTH=YWRtaW46c2VjcmV0
+WPGRAPHQL_VERSION=v1.0.1
+WPGATSBY_VERSION=feature/wpgatsby-stability
```
acao, line 3: this file becomes the control plane for version bump CI logic. enjoy!

TylerBarnes (approved):
Sweet :D this looks great!

PR #364, opened 2020-12-16
ci: make workflows dispatchable

acao:
this allows these workflows to be dispatched remotely via the github api 😮 

also tweaks to expected incoming event payload


TylerBarnes (approved):
🎉

PR #370, opened 2020-12-17
Update action monitor queries to order by date

jasonbahl:
This updates the action monitor queries to order by modified date instead of created date.


TylerBarnes (approved):
Looks great, thank you! :D

PR #371, opened 2020-12-17
Feat: multiple simultaneous preview users

TylerBarnes:
Prior to this PR, if more than 1 user tried to preview at the same time, their preview would hang for a long time (or indefinitely).  In a stress test of just 3 users making 30 previews there was only a 33% success rate. With the work in this PR I've been able to test 10 simultaneous users making 100 previews over a few minutes with a 100% success rate. The upper limit of 10 users so far is only due to running this on my local machine. Soon I'll release a canary and run the stress test script from multiple computers at once against a Gatsby Cloud instance to see what the upper limit is.

The WPGatsby side of this work is here https://github.com/gatsbyjs/wp-gatsby/pull/83


smthomas (approved):

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@6.0.0`

PR #389, opened 2020-12-22
chore: use unicode emojis rather than the github shortcut

gillkyle:
This will make it so the emojis render on other sites like https://www.gatsbyjs.com/plugins/gatsby-source-wordpress-experimental/?=wordpress-exp


TylerBarnes (approved):
Thanks Kyle!

Issue #375, opened 2020-12-18
Error fetching Media with no clear message on where is coming from. 


DanielRiveraHQ:
I’m having an issue with gatsby develop and gatsby build when it pulls assets from WordPress. It seems to be related to a missing asset. Usually, the terminal lets you know if there is a 404 getting an error and the URL, but not this time. Do you guys have a clue? Here is a screenshot of the error:

![Screen Shot 2020-12-17 at 6 27 34 PM](https://user-images.githubusercontent.com/20131939/102633680-5c62d900-4116-11eb-87c2-745fd743fb50.png)

I'm using 
“gatsby-source-WordPress-experimental”: “^5.0.2"
WPGatsby 0.8.0
WP GraphQL 1.0.3

@jasonbahl said he could reproduce the error but not have an answer to fix it and post it here to look into it further. 

In case anyone needs access to the repository, ping me, and I can add you as a collaborator.

cstefanache:
I am facing the same issue and I would like to start working on this. How would you prefer to deal with this issue:
1. Node won't get created and content will ignore the missing content
2. Content will be ignored - page won't be created

DanielRiveraHQ:
Probably just the specific node, and maybe a notification to know where the error is happening. What do you think?

Also, I'm new at this, but if there is any other way I can help, let me know.

cstefanache:
We might need the advice from somebody who maintains this project - this seems to be a big decision

ryancybul:
I am having the same issue as well.  I was able to use the gatsby-source-worpress@v3 plugin to successfully pull the data.  I also tried using gatsby-source-graphql as well and that also worked for me.  For now I'm going to use gatsby-source-graphql and eventually switch over to gatsby-source-worpress-experimental when the kinks are worked out.  In the experimental docs it looks like migrating from gatsby-source-graphql to gatsby-source-worpress-experimental requires less work than migrating from v3.

cstefanache:
with the latest build of  gatsby-source-worpress-experimental and WP-Gatsby 0.9 the issue no longer reproducible.
@DanielRiveraHQ & @ryancybul can you please confirm so I can close the issue?

ryancybul:
[...]

Issue #386, opened 2020-12-20
Wordpress updates are not picked up by gatsby-develop & gatsby-build without gatsby clean


insidesmart:
Up until version 2.1.3, wordpress changes were propagating to gatsby-develop process.

After upgrade to 6.0.0 (gatsby latest version. also upgraded all the required wordpress plugins to the latest), the updates are not reflected in gatsby develop within nodeUpdateInterval time.

Sadly even if i attempt gatsby-develop / gatsby build as a fresh command, the updates are not picked up. If we do gatsby clean and then gataby develop / gatsby build, wordpress updates are picked up as expected.

(Not added any environment variables like ENABLE_GATSBY_REFRESH_ENDPOINT)

brandensilva:
Also experiencing this problem. I will report back with more details when I get a moment.

MuradBuyukasik:
I'm also experiencing this problem. 
`    "gatsby-source-wordpress-experimental": "^6.0.0",
`
WordPress version 5.6,
WP Gatsby v0.9.1
WP Graphql v1.0.3

tj-mc:
Appear to facing this issues also. Can't determine what has changed to trigger this.

TylerBarnes:
Is anyone able to share a reproduction repo?

TylerBarnes:
https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/pull/400 may fix this problem. I identified and fixed an issue there that seemed very similar. I'll post here once that's released so you can try it out :)

TylerBarnes:
I believe this is fixed in `gatsby-source-wordpress-experimental@6.1.0`. Please try it out and let me know if you're still having issues. If you are, we'll need a reproduction to tackle this one. Thanks everyone!

robertschneiderman:
I have this same problem. Upgrading to 6.1.0 doesn't seem to work :/ I upgraded all the way from 1.7.9 => 6.0.0, so I can't give pinpoint when this breaks... but things were working great on 1.7.9

TylerBarnes:
@robertschneiderman are you able to share a reproduction so we can fix it? Thanks :)

mikexavier:
I'm not sure if this is the same issue or not, as it's not necessarily anything to do with updating data on WP.

[...]

Issue #384, opened 2020-12-19
When adding or deleting a media item that has been assigned to a field preview-build erorr


garryburch:
When deleting media items or adding a different media item to a field during dev/preview

```
6:23:03 PM: ⠀
6:23:03 PM: info  gatsby-source-wordpress  deleted mediaItem continuous-analytics (#28464)
6:23:03 PM: ⠀
6:23:03 PM: ⠀
6:23:03 PM: info  gatsby-source-wordpress  deleted mediaItem fast-analytics (#28466)
6:23:03 PM: ⠀
6:23:03 PM: ⠀
6:23:03 PM: info  gatsby-source-wordpress  deleted mediaItem comprehensive-analytics (#28462)
6:23:03 PM: ⠀
6:23:03 PM: ⠀
6:23:03 PM: info  gatsby-source-wordpress  update page Free Trial Landing Page (#22212)
6:23:03 PM: ⠀
6:23:03 PM: ⠀
6:23:04 PM: ⠀
6:23:04 PM: info  gatsby-source-wordpress  update page KPI Diagnosis Landing Page (#21071)
6:23:04 PM: ⠀
6:23:04 PM: ⠀
6:23:04 PM: ⠀
6:23:04 PM: info  gatsby-source-wordpress  update page Request A Demo (#1024)
6:23:04 PM: ⠀
6:23:04 PM: ⠀
6:23:32 PM: error Error: url passed to create-remote-file-node is either missing or not a proper web uri: /static/cb56d536ef0694c2e829c2889b3b2924/continuous-analytics-v2.svg
6:23:32 PM:     at module.exports (/opt/build/repo/node_modules/gatsby-source-wordpress-experimental/src/steps/source-nodes/create-nodes/create-remote-file-node/index.js:427:7)
6:23:32 PM:     at remoteFileNode (/opt/build/repo/node_modules/gatsby-source-wordpress-experimental/src/steps/source-nodes/create-nodes/create-remote-media-item-node.js:268:26)
6:23:32 PM:     at RetryOperation.runAttempt [as _fn] (/opt/build/repo/node_modules/async-retry/lib/index.js:41:15)
6:23:32 PM:     at Timeout._onTimeout (/opt/build/repo/node_modules/retry/lib/retry_operation.js:81:10)
6:23:32 PM:     at ontimeout (timers.js:436:11)
6:23:32 PM:     at tryOnTimeout (timers.js:300:5)
[...]

PR #400, opened 2021-01-07
Fix: non node images fail with no error message

TylerBarnes:
### Bug Fixes & internal changes

- Since we unified the code for fetching media item nodes and file nodes in html fields, a regression was introduced where our http error handling was assuming it was still always for a media item node. This oversight obscured errors for html image file sourcing.
- Sort actions by modified date instead of date since there's now only 1 action for each node which is just modified instead of being created new.
- The last sourced time was moved from the end of sourcing data to the beginning. The reason for this is that if updates happen during node sourcing, they could previously be potentially missed. Now they'll be picked up.
- Static file link replacement could previously be triggered on MediaItem nodes. This was problematic because the sourceUrl/mediaItemUrl field would be overwritten with a static file link. When the plugin later went to use these fields to create the localFile field, it was no longer a valid url for the WP site and the build would crash. Static file link replacement is now disabled for MediaItem nodes to prevent this from happening.
- Removed dead code which was keeping track of actions and removing duplicates. We now only ever store 1 action per node so duplicates aren't possible anymore.

### New Features

- Because 404's can block many sites from completing production builds, we've added an option to allow 404s to not fail production builds.

```
{
  options: {
    production: {
      allow404Images: true
    }
  }
}
```


TylerBarnes:
Closes #375, #206, and #384

plugin/src/steps/source-nodes/update-nodes/content-update-interval.js
```diff
@@ -11,7 +11,8 @@ import { LAST_COMPLETED_SOURCE_TIME } from "../../../constants"
  */
 const checkForNodeUpdates = async ({ cache, emitter }) => {
   // get the last sourced time
-  const since = await cache.get(LAST_COMPLETED_SOURCE_TIME)
+  const lastCompletedSourceTime = await cache.get(LAST_COMPLETED_SOURCE_TIME)
+  const since = lastCompletedSourceTime - 500
```
smthomas, line 15: Where does the 500 value come from here?

TylerBarnes: It's somewhat arbitrary - often we were missing node updates by a few milliseconds during development. It has no effect on production builds but makes development content updates reliable. We actually had this in here for a really long time and I removed it before the holidays when Jason did his big refactor. I thought the refactor had removed the need for this but it's still an issue.

TylerBarnes: If we switched from using a timestamp to a cursor this wouldn't be an issue anymore, but that's too big of a change right now :)

TylerBarnes: This is the problem -> https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/issues/386

smthomas (approved):

jasonbahl (approved):

TylerBarnes:
Published in `gatsby-source-wordpress-experimental@6.1.0`

PR #405, opened 2021-01-07
Fix: delete menu items when menus are deleted (and more)

TylerBarnes:
Originally this PR was supposed to just be for deleting menu items when their parent menu was deleted, but I uncovered a few bugs along the way that needed to be fixed for this to work properly:

- GraphQL resolvers were previously attempting to lazy fetch nodes if the node isn't found in Gatsby. This is undesired behaviour because it will slow down resolvers considerably for any connections to deleted nodes. This behaviour was originally only intended for MediaItem nodes which could be set to be fetch lazily in resolvers. The fix here is to scope this behaviour only to MediaItem nodes and only when the MediaItem.lazyNodes option is enabled.
- Due to a WPGraphQL bug where menu items can have a different id depending on which entry point they were fetched from, we had a problem of duplicated menu item nodes in some cases. This is temporarily fixed in this plugin until it's fixed upstream.
- When deleting a menu in WP, all child menu items are deleted. We weren't previously accounting for this which meant there would be floater menu items in Gatsby which had no parent menu when a menu was deleted. We now clean up these menu item nodes automatically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## Upcoming
+
+### Bug Fixes
+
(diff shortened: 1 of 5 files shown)


smthomas (approved):

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@6.1.1`

PR #409, opened 2021-01-09
Feat: preview debug mode

TylerBarnes:
### Bug Fixes

- Restored lazy nodes resolvers for all nodes in previews. Because of the default preview preset, removing lazy nodes in the last version from everything except media item nodes ended up breaking previews that previously relied on things working this way. So in preview mode lazy nodes in resolvers is restored.
- In Previews when the schema changed all nodes were being refetched. This is only desireable when the code also changes, because previews have no way to automatically adapt to schema changes. Schema updates in previews no longer refetch all nodes.

### New Features

- Added a preview debug mode which outputs relevant information for debugging Previews.
- Added a debug option to print the diff of the previous and next schema to the terminal output when the remote schema changes. This option is automatically enabled in preview debug mode.

From the plugin options docs:

#### debug.graphql.printIntrospectionDiff: Boolean

When this is set to true it will print out the diff between types in the previous and new schema when the schema changes. This is enabled by default when `debug.preview` is enabled.

Default is `false`.

```js
{
  resolve: `gatsby-source-wordpress-experimental`,
  options: {
    debug: {
      graphql: {
        printIntrospectionDiff: true,
      },
    },
  },
},
```

and:

### debug.preview: Boolean

When set to true, this option will display additional information in the terminal output about the running preview process.

Default is `false`.

```js
{
  resolve: `gatsby-source-wordpress-experimental`,
  options: {
    debug: {
      preview: true
    },
  },
},
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 - Restored lazy nodes resolvers for all nodes in previews. Because of the default preview preset, removing lazy nodes in the last version from everything except media item nodes ended up breaking previews that previously relied on things working this way. So in preview mode lazy nodes in resolvers is restored.
 - In Previews when the schema changed all nodes were being refetched. This is only desireable when the code also changes, because previews have no way to automatically adapt to schema changes. Schema updates in previews no longer refetch all nodes.
 
+### New Features
+
+- Added a preview debug mode which outputs relevant information for debugging Previews.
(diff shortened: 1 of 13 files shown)


TylerBarnes:
Example of the schema diff output:

Screen Shot 2021-01-08 at 6 21 23 PM
Screen Shot 2021-01-08 at 6 21 45 PM

smthomas (dismissed):
This is awesome. Nice work.

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@6.2.0`

PR #412, opened 2021-01-11
fix: remove old menu logic that's not needed anymore

TylerBarnes:
### Breaking Changes

- There was a longstanding menu/menu-item bug which is fixed in the latest WPGQL. This release removes the code that was compensating for this bug. This will speed up menu sourcing and reduce the complexity of our codebase. This is a breaking change because it requires that we set the minimum version of WPGQL to the latest version as of this release.

### Bug fixes

- When cleaning up leftover callbacks that send preview status back to WP Previews, we were not including a context message to display in the terminal and in browser when a page wasn't being created for the node being previewed.
- Menu items were not being cleaned up properly after a menu was deleted. We now delete them along with their parent menu.


plugin/src/supported-remote-plugin-versions.ts
```diff
@@ -2,7 +2,7 @@
 // it indicates which versions we will actually support AND which versions work.
 const supportedWpPluginVersions = {
   WPGraphQL: {
-    version: `>=0.14.0 <2.0.0`,
+    version: `>=1.1.2 <2.0.0`,
```
jasonbahl, line 5: nice. force folks to bump to WPGraphQL 1.0+ 💪

TylerBarnes:
I tweaked the compatibility api message too since folks will see it after this release and I know a lot of people find it confusing:

Screen Shot 2021-01-12 at 11 12 34 AM

plugin/src/steps/check-plugin-requirements.ts
```diff
@@ -130,16 +130,16 @@ const areRemotePluginVersionsSatisfied = async ({
       supportedWpPluginVersions.WPGraphQL.version
     )
 
-    message += `Your remote version of WPGraphQL is not within the accepted range (${
+    message += `Your remote version of WPGraphQL is not within the accepted range\n(${
       supportedWpPluginVersions.WPGraphQL.version
-    }).
+    }).\n\nThis is not a bug and it means one of two things:\n you either need to upgrade either WPGraphQL or gatsby-source-wordpress-experimental.
```
jasonbahl: I like these updates. a bit more clarity

jasonbahl (dismissed):

jasonbahl:
LGTM 🐴

TylerBarnes:
This is published in `gatsby-source-wordpress-experimental@7.0.0`

PR #415, opened 2021-01-15
Fix: account for stale previews

TylerBarnes:
### Changes

- Lowered `options.schema.previewRequestConcurrency` to 5.
- Added a sinceTimestamp where arg to preview sourcing so that only previews created in the last 10 minutes are sourced.


jasonbahl (approved):

jasonbahl:
LGTM 🐴

PR #417, opened 2021-01-16
Fix: don't access property on undefined when printing schema diff

TylerBarnes:
### Bug Fixes

- The new schema diff out debugging api could fail when accessing a property on undefined in certain cases.


TylerBarnes:
Published in `gatsby-source-wordpress-experimental@7.0.3`

PR #424, opened 2021-01-20
Feat: generate plugin options docs

TylerBarnes:
This PR adds a script which generates plugin options docs.


TylerBarnes:
This is ready for review!

btw @mxstbr you might find this interesting. I'm using your plugin options schema API to generate our plugin options documentation https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/8f51f075b44ee22564cf94c599454dd25ea75eda/docs/plugin-options.md

It used to be manually maintained 😱 
https://github.com/gatsbyjs/gatsby-source-wordpress-experimental/blob/master/docs/plugin-options.md

jasonbahl (approved):
@TylerBarnes this looks 🔥 

Only thing I think we should follow up with (whenever) is to automate this. Any PR where the plugin options change should maybe re-generate the plugin options automatically? 

The CI part can be addressed anytime though.

TylerBarnes:
Thanks @jasonbahl and good call on automation. I left it out for now because I'm not 100% sure what that'll look like in the Gatsby monorepo.

jasonbahl:
Ah ya. . .monorepo 😬

PR #433, opened 2021-01-26
Fix: broken plugin options schema

TylerBarnes:
Well this was a hard one to figure out. Gatsby is using the deprecated version of Joi for which there are no docs. I had changed out plugin options schema to work with the latest joi version (not realizing Gatsby isn't using this) so I could generate our plugin options.. well this resulted in Gatsby swallowing the joi error somehow and the build crashing on something completely unrelated... I finally figured out what was causing it by trial and error (it was a default value in the joi schema that was a string where the actual field is not a string and latest joi allows examples that are strings where the deprecated Gatsby version only allows examples that are of the type of the field...

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "@babel/core": "^7.6.4",
     "@babel/plugin-proposal-class-properties": "^7.10.1",
     "@babel/plugin-proposal-private-methods": "^7.10.1",
-    "@hapi/joi": "^17.1.1",
+    "@hapi/joi": "^15.1.1",
     "@typescript-eslint/eslint-plugin": "^4.5.0",
     "@typescript-eslint/parser": "^4.5.0",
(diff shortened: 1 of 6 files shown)


jasonbahl:
Should rename this PR something punny. Like "oh the joi of silent failures"

package.json
```diff
@@ -44,7 +44,7 @@
     "@babel/core": "^7.6.4",
     "@babel/plugin-proposal-class-properties": "^7.10.1",
     "@babel/plugin-proposal-private-methods": "^7.10.1",
-    "@hapi/joi": "^17.1.1",
+    "@hapi/joi": "^15.1.1",
```
jasonbahl, line 47: dope. Love moving to deprecated dependencies! 💪

jasonbahl (approved):

PR #438, opened 2021-01-28
Docs: add transforming data tutorial

TylerBarnes:
This mini tutorial explains how to transform node data on its way to Gatsby. This tutorial is intentionally brief as it's only meant for advanced developers who are building Gatsby plugins or very much know what they're getting into.

### DIFF
--- a/docs/tutorials/transforming-data.md
+++ b/docs/tutorials/transforming-data.md
@@ -8,14 +8,10 @@ A good example on how to use this API is by referencing internal uses of it. In 
 
 ```js
 const defaultPluginOptions = {
-    // ...
+  // ...
   type: {
     MediaItem: {
-      beforeChangeNode: async ({
-        remoteNode,
-        actionType,
-        typeSettings,
-      }) => {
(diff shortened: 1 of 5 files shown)


docs/tutorials/transforming-data.md
```diff
@@ -0,0 +1,106 @@
...
+        node: menuItemNode,
+      })
+    )
+  }
+}
+```
+
+Here we're finding all child MenuItem's on the MenuItem being deleted and then deleting those items. We do this as a performance optimization. In WP if a Menu is deleted, the child MenuItem's are also deleted. Instead of sending an event for each item and needing Gatsby to fetch all of them, we know that when we receive an event to delete a Menu, we can safely delete its child MenuItem's.
+
+# Cancelling an update
```
jasonbahl, line 89: I believe this should be an h2 `##`? or did you intend it to be another H1?

jasonbahl (changes requested):
one change, then LGTM 🐴

jasonbahl (approved):

PR #453, opened 2021-02-02
Fix url encoding when fetching remote assets

therealgilles:
Use url.parse(url).href instead of decodeURI/encodeURI to properly encode the url if necessary.
This fixes the issue where an avatar url was getting double-encoded.


therealgilles:
Oops, how did I not see the conflict between the url package and the url var, fix coming...

therealgilles:
I used `const urlParse = require('url').parse` to avoid the conflict with existing `url` variables. Let me know if it's undesirable. If so, feel free to suggest an alternative names for either the library variable or the existing `url` variables.

